PR #59, opened 2024-07-09
refactor: enhance error handling with custom ConfigError type

codope:
- Created ConfigError enum for more robust error handling
- Updated ConfigParser trait and its implementations to use ConfigError
- Modified HudiInternalConfig, HudiReadConfig, and HudiTableConfig to leverage ConfigError
- Adjusted tests to align with new error handling

### DIFF
--- a/crates/core/src/config/mod.rs
+++ b/crates/core/src/config/mod.rs
@@ -18,13 +18,32 @@
  */
 use std::any::type_name;
 use std::collections::HashMap;
+use std::error::Error;
+use std::fmt;
 use std::sync::Arc;
-
-use anyhow::Result;
 
 pub mod internal;
 pub mod read;
 pub mod table;
+
+#[derive(Debug)]
+pub enum ConfigError {
+    NotFound,
+    ParseError(String),
+    Other(String),
+}
+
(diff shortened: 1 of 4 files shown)


crates/core/src/config/mod.rs
```diff
@@ -18,14 +18,33 @@
...
-
 pub mod internal;
 pub mod read;
 pub mod table;
 
+#[derive(Debug)]
+pub enum ConfigError {
+    NotFound,
+    ParseError(String),
+    Other(String),
```
xushiyan, line 33: that's a nice improvement. I would suggest capture underlying error as source, like ParseError should capture std::ParseIntError, etc, and NotFound should capture which key (ConfigParser) it refers to.

On a bigger scope, we should definitely standardize error types throughout hudi-core and other hudi crates. I chose `anyhow` for fast iteration and uncover error handling paths first. So all errors come out from hudi are now anyhow::Error. I suggest replace anyhow dependency with well-defined custom error enums implemented with [thiserror](https://docs.rs/thiserror/latest/thiserror/) in the next release.

xushiyan:
this was covered in https://github.com/apache/hudi-rs/pull/215

Issue #43, opened 2024-07-05
Define labels and setup auto labeling


xushiyan:
https://github.com/actions/labeler

PR #83, opened 2024-07-19
chore: configure labeler for PRs from forked repos

xushiyan:
For #43


xushiyan:
as per https://github.com/actions/labeler?tab=readme-ov-file#notes-regarding-pull_request_target-event, we need to land this first then adjust config if needed.

PR #93, opened 2024-07-24
ci: add path ignore files for ci

### LINKED ISSUE #84: Configure workflows to avoid running tests when changes are not relevant

### LINKED ISSUE #91: Setup dependabot as per ASF guide
Ref: https://github.com/apache/infrastructure-asfyaml

abyssnlp:
## Description

- Only run tests when a change is made on any of the relevant paths.
- Move license and style checks to another CI workflow

Closes #84 

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,16 +23,28 @@ on:
       - main
       - release/**
     paths:
+      - 'Cargo.toml'
+      - 'codecov.yml'
+      - 'rust-toolchain.toml'
       - 'python/**'
-      - '!python/Makefile'
       - 'crates/**'
+    paths-ignore:
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/PULL_REQUEST_TEMPLATE.md'
   pull_request:
     branches:
       - main
     paths:
+      - 'Cargo.toml'
+      - 'codecov.yml'
+      - 'rust-toolchain.toml'
       - 'python/**'
-      - '!python/Makefile'
       - 'crates/**'
+    paths-ignore:
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/PULL_REQUEST_TEMPLATE.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -22,9 +22,17 @@ on:
     branches:
       - main
       - release/**
+    paths:
+      - 'python/**'
+      - '!python/Makefile'
+      - 'crates/**'
```
xushiyan, line 28: we shouldn't exclude Makefile as some CI jobs may make use of the commands there. code and project setup/dependencies related files should be CI-covered. some example files that can be exempted are like README, NOTICE, LICENSE, etc.

abyssnlp: Addressed in [174894c](https://github.com/apache/hudi-rs/pull/93/commits/174894c1d22b9de9fc3d4d1786859c9a5c65062a). The CI should now run whenever a change is made in either the rust or python source, setup/project dependencies or changes in test coverage configuration. Although redundant as the paths spec should take care of this, I've explicitly ignored running CI on paths for markdown files and changes in the issue template or pull request template.

xushiyan: hmmm this causes maintenance burden as project evolves - we have to keep this list updated so that CI will trigger, note that we already have crates-level Cargo.toml, what about pyproject.toml? wouldn't just defining `paths-ignore` serve the purpose?

abyssnlp: Fair point. My reasoning was that as the project evolves, majority of the changes would be within the `crates/` and `python/` paths. The trade-off here would be between maintaining a list of paths to trigger the CI on vs. maintaining the list of paths to ignore.
But if non-project dependencies won't change as much, I agree it would be easier to just ignore those instead. I'll make the required changes.

abyssnlp: Does it make sense to have `cliff.toml` in there as its used to generate changelogs so would be irrelevant for tests?

xushiyan: we might need CI to auto generate changelog in future so let's keep it out

abyssnlp: Please feel free to resolve this and proceed if everything looks okay.

xushiyan (approved):

PR #112, opened 2024-08-13
test: add integration test with minio

### LINKED ISSUE #81: Setup integration test with minio

abyssnlp:
## Description

Adds integration tests with s3-compatible MinIO.
- `docker/docker-compose.yaml` spins up the minio container
- `docker/copy_tables.sh` creates the bucket; unzips, copies and creates the tables in the minio store
- The environment variables for the integration test are supplied via `integration_test/env.toml`
- Adds running integration tests on pull requests via `ci.yml`

### Notes:
- We can extract the logic to run the integration tests in the `Makefile` or as a separate script
- The integration test takes around ~3 minutes to complete
- The integration test is only for the `hudi-rs` rust support via datafusion

This closes #81 .

- [ ] N/A
- [ ] Manual tests


xushiyan:
@abyssnlp don't think the ci job is triggered. probably format issue.

abyssnlp:
@xushiyan  The CI is failing due to this error: 
`dtolnay/rust-toolchain@1.75.0 is not allowed to be used in apache/hudi-rs.` I used this action to setup the rust toolchain. Is there a way to whitelist this action or should I use an alternative?

abyssnlp:
I've removed the external action since GHA runner has `rustup` already.

abyssnlp:
@xushiyan All tests (including integration) are green. The labeler is failing due to [permissions](https://github.com/actions/labeler#permissions).

integration_test/tests/hudi_test.rs
```diff
@@ -0,0 +1,276 @@
...
+#[derive(PartialEq, Debug)]
+struct Record {
+    id: i32,
+    name: String,
+    is_active: bool,
+}
+
+#[derive(Debug, EnumString, EnumIter, AsRefStr, PartialEq)]
+#[strum(serialize_all = "snake_case")]
+pub enum IntegrationTestTable {
```
xushiyan, line 44: can you reuse the code from crates/tests/src/lib.rs:39

integration_test/tests/hudi_test.rs
```diff
@@ -0,0 +1,276 @@
...
+    V6ComplexkeygenHivestyle,
+    V6Empty,
+    V6Nonpartitioned,
+    V6SimplekeygenHivestyleNoMetafields,
+    V6SimplekeygenNonhivestyle,
+    V6SimplekeygenNonhivestyleOverwritetable, // 1, 21
```
xushiyan, line 50: what does 1, 21 mean?

integration_test/tests/hudi_test.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    pub fn url(&self) -> Url {
+        let path = self.path();
+        Url::parse(&path).unwrap()
+    }
+}
+
+#[test]
+fn test_table_path() {
```
xushiyan, line 69: in integ test we can leave out this path testing

.github/workflows/pr.yml
```diff
@@ -18,7 +18,7 @@
 name: PR
 
 on:
-  pull_request:
+  pull_request_target:
```
xushiyan, line 21: let's revert this. there is asf policy against using this. we can ignore the labeler issue for now.

abyssnlp:
@xushiyan Addressed all comments in [a28debc](https://github.com/apache/hudi-rs/pull/112/commits/a28debc252bae3d9c8334d5a9aef4a35b66aa5a7). 

I've reused the `TestTable` enum but I had to add an extra method to get the s3 path as we unzip and load the tables into minio for the integration tests so we don't need to extract them.

xushiyan (commented):
thanks for putting this up! Some high-level feedback

- How we plan for `docker/`: we can position it as a sandbox where people can easily spin up some e2e setup and run something, we may want to add more query engines later, so integration test can be part of it, serving as a validation to make sure all components are working.
- Right now the docker setup is just providing minio, and the integ test is running natively and connecting to minio container. To align with the plan above, we should move the test into a container itself, which installs crates from the code repo.
- The integ test should also cover a python test case by pip install hudi from the repo

xushiyan:
@abyssnlp are you planning to resume this work?

abyssnlp:
Hi @xushiyan, 

I've been busier than usual so couldn't get around to this earlier. I've addressed your comments in https://github.com/apache/hudi-rs/pull/112/commits/5a747fd1dc45fccf85cfdcad6fd8ae5068fc8993.

- The integration tests now run in a container
- I'll add a test for python as well later

Let me know what you think.

xushiyan:
Ok, due to time constraint, I'll push this to the next release.

xushiyan:
re-worked in https://github.com/apache/hudi-rs/pull/226

PR #114, opened 2024-08-18
chore: fix has no write access from a forked repository when add label

KnightChess:
## Description

has no write access from a forked repository when add label.
image

as github advice https://github.com/actions/labeler?tab=readme-ov-file#permissions, use `pull_request_target`.

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,7 +18,7 @@
 name: PR
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


xushiyan:
@KnightChess thanks. but i've been checking with asf infra as there is policy problem. we can disable this for now https://github.com/apache/hudi-rs/pull/115

Issue #47, opened 2024-07-05
Support partition pruning during file listing


xushiyan:
Provide basic table API to accept predicates like `foo > 10` or `bar != A`. The partition loading for file system view should process the predicates and load relevant partitions. Just consider supporting `AND` for multiple predicate string expressions.

Follow up work #159 #160

KnightChess:
Hi, I'd like to work for this feature if no one plan work for it.

xushiyan:
@KnightChess great! can you please describe a high-level approach for the implementation here?

KnightChess:
support hudi internal filter, engine like datafusion or other need cover it expression to hudi partition filter.
something like this code, 
```rust 
let filter_one = PartitionFilter::try_from(("shortField", "=", "100", short_field_data_type)).unwrap();
let filter_two = PartitionFilter::try_from(("shortField", ">", "100", short_field_data_type)).unwrap();
hudi_table.partition_filter_replace(vec![filter_one, filter_two]);
```

the core struct filter define, key is partirion field, value is expression-value
```rust 
pub struct PartitionFilter {
    /// The key of the PartitionFilter
    pub key: String,
    /// The value of the PartitionFilter
    pub value: PartitionValue
}
```
expression-value: reuse datafusion `ScalarValue`.
```rust 
pub enum PartitionValue {
    /// The partition value with the equal operator
    Equal(ScalarValue),
    /// The partition value with the not equal operator
    NotEqual(ScalarValue),
    /// The partition value with the greater than operator
    GreaterThan(ScalarValue),
    /// The partition value with the greater than or equal operator
    GreaterThanOrEqual(ScalarValue),
    /// The partition value with the less than operator
    LessThan(ScalarValue),
    /// The partition value with the less than or equal operator
    LessThanOrEqual(ScalarValue),
    /// The partition values with the in operator
    In(Vec<ScalarValue>),
    /// The partition values with the not in operator
    NotIn(Vec<ScalarValue>),
}

```

PR #119, opened 2024-08-25
feat: support partition prune api

KnightChess:
## Description

Add filtering capabilities to table API, currently only partition fields are applicable. Multiple predicates are `AND` together.

```rust
hudi_table.read_snapshot(&["foo != a", "bar >= 100"]);
```
Supported operators are: `>, >=, <, <=, =, !=`.

For #47 

- [ ] N/A
- [ ] Manual tests


KnightChess:
@xushiyan cc

KnightChess:
@xushiyan hello, Is there a problem with my test method in my local?
image
image

KnightChess:
> hello, Is there a problem with my test method in my local?
ignore checkstyle, fetch the latest commit, it work again, but can not report python test error.

KnightChess:
@xushiyan cc:

xushiyan:
@KnightChess awesome contribution! let me take a look. i might push some quick fixes just FYI to move faster.

crates/core/Cargo.toml
```diff
@@ -63,8 +63,9 @@ tokio = { workspace = true }
 # datafusion
 datafusion = { workspace = true, optional = true }
 datafusion-expr = { workspace = true, optional = true }
-datafusion-common = { workspace = true, optional = true }
+datafusion-common = { workspace = true }
 datafusion-physical-expr = { workspace = true, optional = true }
+log = "0.4.22"
```
xushiyan, line 68: can this dep be added to workspace for centralized management?

crates/core/Cargo.toml
```diff
@@ -63,8 +63,9 @@ tokio = { workspace = true }
 # datafusion
 datafusion = { workspace = true, optional = true }
 datafusion-expr = { workspace = true, optional = true }
-datafusion-common = { workspace = true, optional = true }
+datafusion-common = { workspace = true }
```
xushiyan, line 66: hudi-core should be optionally dependent on datafusion - only when datafusion feature is enabled. let's keep this optional

crates/core/src/table/partitions.rs
```diff
@@ -0,0 +1,362 @@
...
+use std::collections::HashMap;
+
+/// A special value used in Hive to represent the null partition in partitioned tables
+pub const NULL_PARTITION_VALUE_DATA_PATH: &str = "__HIVE_DEFAULT_PARTITION__";
+
+/// A Enum used for selecting the partition value operation when filtering a HudiTable partition.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PartitionValue {
+    /// The partition value with the equal operator
+    Equal(ScalarValue),
```
xushiyan, line 34: echoing comment at the beginning - can we remove the dependency of ScalarValue from datafusion? wouldn't we be ok to keep strings in the model and convert using DataType for comparison when matching partitions? use something from arrow if suitable

crates/core/src/table/fs_view.rs
```diff
@@ -70,7 +73,31 @@ impl FileSystemView {
...
+        if partition_filters.is_empty() {
+            return Ok(partition_paths);
+        }
+        let field_and_data_type: HashMap<_, _> = partition_schema
+            .fields()
+            .iter()
+            .map(|field| (field.name().to_string(), field.data_type().clone()))
+            .collect();
+
+        Ok(partition_paths
```
xushiyan, line 85: all partition paths were listed and then filtered - a more optimized approach is to filter out partitions during the listing process, will reduce a lot of scanning for multi-level partitions. we can optimize this in a separate PR

xushiyan (changes requested):
@KnightChess good progress. i think the main functionalities are there. A major change is to re-model the Partition struct, we need to decouple it from datafusion struct

crates/core/src/table/mod.rs
```diff
@@ -234,6 +239,23 @@ impl Table {
...
 
+    pub async fn get_partition_schema(&self) -> Result<Schema> {
+        let schema = self.get_schema().await;
+        let partition_str_vec = self
+            .configs
+            .get_or_default(PartitionFields)
+            .to::<Vec<String>>();
+        let partition_set: HashSet<String> = partition_str_vec.into_iter().collect();
+        let partition_fields: Vec<Arc<Field>> = schema
+            .unwrap()
```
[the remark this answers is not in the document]
xushiyan: this was fixed.

crates/core/src/table/mod.rs
```diff
@@ -234,6 +239,23 @@ impl Table {
         self.timeline.get_latest_schema().await
     }
 
+    pub async fn get_partition_schema(&self) -> Result<Schema> {
+        let schema = self.get_schema().await;
+        let partition_str_vec = self
```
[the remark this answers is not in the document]
xushiyan: this was fixed.

crates/core/src/table/mod.rs
```diff
@@ -306,6 +330,16 @@ impl Table {
...
+
+    pub fn partition_filter_replace(&mut self, partition_filters: Vec<PartitionFilter>) {
+        if self.configs.get_or_default(IsHiveStylePartitioning).to() {
+            self.partition_filters = partition_filters;
+        }
+    }
+
+    pub fn reset(&mut self) {
+        self.file_system_view.reset()
+    }
```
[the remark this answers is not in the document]
xushiyan: @KnightChess in this PR we can annotate them as test helper like `#[cfg(test)]`, then you can file a separate PR to refactor the test so we won't allow mutability through these table API.

xushiyan:
@KnightChess do you think you can address the main comment in the next few days? then i can polish further if needed and land this. Trying to get this in the upcoming release within 2 weeks 🙂

KnightChess:
@xushiyan sorry for reply late, I will address these two days

crates/core/src/table/mod.rs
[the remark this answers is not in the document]
KnightChess: @xushiyan This is not only for testing, but also to refresh the cache, like spark need use `refresth table defualt.testA` to refresth cache meta.

crates/core/Cargo.toml
KnightChess: oh, forgot it, will update it.

crates/core/Cargo.toml
KnightChess: It now relies on the `ScalarValue` of DataFusion.

crates/core/src/table/partitions.rs
KnightChess: thanks, I will try to remove `the dependency of ScalarValue` and fix `datafusion-common = { workspace = true }` be optional.

crates/core/src/table/fs_view.rs
KnightChess: good idea, will create a separate PR improve it.

crates/core/src/table/mod.rs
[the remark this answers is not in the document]
xushiyan: @KnightChess yes understood, eventually we'll need to implement those functionalities. But let's stick to the on-demand principle - don't "predict" the needs, adding APIs when it becomes necessary. As of the codebase, these 2 serve for testing so let's write proper test code, and we can evolve them for production code whenever need to.

KnightChess:
@xushiyan Hello, I couldn't find an implementation similar to ScalarValue, and I am not very familiar with Arrow yet, and. There is a certain learning curve involved, which might delay the progress of this PR. Could you please help improve this PR?

KnightChess:
@xushiyan cc, I try to use arrow `Scalar<ArrayRef>` to replace datafusion `ScalarValue`, and modified some suggestions to repair.

crates/core/src/config/table.rs
```diff
@@ -130,6 +130,7 @@ impl ConfigParser for HudiTableConfig {
         match self {
             Self::DatabaseName => Some(HudiConfigValue::String("default".to_string())),
             Self::DropsPartitionFields => Some(HudiConfigValue::Boolean(false)),
+            Self::PartitionFields => Some(HudiConfigValue::List(vec!["".to_string()])),
```
xushiyan, line 133: default as a list of an empty str ? should it be an empty list

crates/core/src/table/partitions.rs
```diff
@@ -289,59 +267,6 @@ impl TryFrom<(&str, &HashMap<String, DataType>)> for HudiTablePartition {
...
 mod tests {
     use super::*;
-    use serde_json::json;
-
-    fn check_json_serialize(filter: PartitionFilter, expected_json: &str) {
-        assert_eq!(serde_json::to_value(filter).unwrap(), json!(expected_json))
-    }
-
-    #[test]
-    fn test_serialize_partition_filter() {
```
xushiyan: @KnightChess why the serialization test was removed?

KnightChess: @xushiyan arrow scalar not implement `Display`, and I found the scenes that are not used at present, so I remove `Serialize` trait. If necessary, I can add implementation

crates/core/Cargo.toml
```diff
@@ -73,6 +74,5 @@ hudi-tests = { path = "../tests" }
 datafusion = [
     "dep:datafusion",
     "datafusion-expr",
-    "datafusion-common",
```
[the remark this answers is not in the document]
KnightChess: I forgot to add it back. It seems that the test has a issue, and there is no error without this dependence.

crates/core/src/table/partitions.rs
KnightChess: @xushiyan thank you so much

xushiyan:
I was wrapping up my vacation 😄  just now getting back to update this:

Changes I've made:
- Add `PartitionPruner` to handle pruning logic, taking care of hudi configs like hive style and url encoded paths.
- Taking filters from parameters in form of `foo > 10`, `bar != a`, etc, only apply `AND` to them.
- Limit the operators to only `>, >=, <, <=, =, !=` to avoid rabbit hole of implementing sql parser; `IN` and `NOT IN` can be achieved anyway.
- API to reset should not be needed now, as the pruning with different filters should be applied to loaded fs view directly.

There are more follow up work to do on datafusion integration side, which I'll jot down in the GH issue.

xushiyan (approved):

KnightChess:
@xushiyan thanks review

PR #122, opened 2024-08-30
docs: add the examples crate for rust

gohalo:

- [x] N/A
- [ ] Manual tests

### DIFF
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -1,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "hudi-examples"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+rust-version.workspace = true
+keywords.workspace = true
+readme.workspace = true
+description.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+[dependencies]
+hudi = { path = "../hudi", features=["datafusion"] }
+
+[dev-dependencies]
+tokio = { workspace = true, features=["macros"] }
+datafusion = { workspace = true }
+
+[[example]]
+name = "hello"
+path = "hello.rs"
--- a/crates/examples/hello.rs
+++ b/crates/examples/hello.rs
@@ -1,0 +1,20 @@
+use std::sync::Arc;
+use std::env::var;
+
+use datafusion::{
+    error::Result,
+    prelude::{DataFrame, SessionContext},
+};
+use hudi::HudiDataSource;
+
+#[tokio::main]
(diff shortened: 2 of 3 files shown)


crates/examples/hello.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use datafusion::{
+    error::Result,
+    prelude::{DataFrame, SessionContext},
+};
+use hudi::HudiDataSource;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let data_path = "/your/hudi/data/path";
```
codope, line 11: `data_path` variable should ideally not be hardcoded. Instead, you might want to pass it as a command-line argument or environment variable, allowing the example to be more flexible and reusable in different environments. Something like below:
```
data_path = std::env::var("HUDI_DATA_PATH").unwrap_or_else(|_| "/your/hudi/data/path".to_string());
```

crates/examples/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+description.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+[dependencies]
+hudi = { path = "../hudi", features=["datafusion"] }
+
+[dev-dependencies]
+tokio = { workspace = true, features=["full"] }
+datafusion = { workspace = true }
```
codope, line 35: @xushiyan Could you confirm the dependencies are correct?

codope (commented):
Thanks for adding an example. I am assuming you plan to build more on top of this.

@xushiyan Could you please review as well?

xushiyan:
@gohalo thanks for raising this. While an "examples" crate can be helpful, we won't merge nice-to-have PRs until we can justify the value of the code. As of now, it's a simple example repeating the one on the readme. What other examples could we add there in future? How do we expect users to use and benefit from it? How can the example code be tested properly so we can easily evolve it? If we are to create a website, how can we consolidate the code here and the examples to show on website?

In short, I'm highlighting that we need to make a proper plan or design for any major addition to the codebase.

xushiyan (changes requested):
@gohalo thanks for raising this. While an "examples" crate can be helpful, we won't merge nice-to-have PRs until we can justify the value of the code. As of now, it's a simple example repeating the one on the readme. What other examples could we add there in future? How do we expect users to use and benefit from it? How can the example code be tested properly so we can easily evolve it? If we are to create a website, how can we consolidate the code here and the examples to show on website?

In short, I'm highlighting that we need to make a proper plan or design for any major addition to the codebase. You may want to create an issue for design, plan, and discussion first.

gohalo:
> @gohalo thanks for raising this. While an "examples" crate can be helpful, we won't merge nice-to-have PRs until we can justify the value of the code. As of now, it's a simple example repeating the one on the readme. What other examples could we add there in future? How do we expect users to use and benefit from it? How can the example code be tested properly so we can easily evolve it? If we are to create a website, how can we consolidate the code here and the examples to show on website?

It's just a startup example for now. Recently, i'm trying to build a local minio env, and also working on create the table, support writing and so on. Maybe the examples will be a good startup for user and especilly for developer to run some basic example.
> 
> In short, I'm highlighting that we need to make a proper plan or design for any major addition to the codebase. You may want to create an issue for design, plan, and discussion first.

Just open an issue #123

crates/examples/Cargo.toml
gohalo: maybe only the macros features will work.

xushiyan:
we should evolve the changes such that the examples are embedded in the rust module level docs and shown in docs.rs like https://docs.rs/pyo3/latest/pyo3/#example-using-python-from-rust
this is more canonical way of showing examples, and automatically we'll get the example code tested by doc testing.

xushiyan:
Thanks for the PR. While we're adding module level examples like https://github.com/apache/hudi-rs/blob/4176c2966ad3042aa4ca49dc5f231cf58a613d8f/crates/core/src/table/mod.rs#L24

we also have demo apps that ran as integ tests: https://github.com/apache/hudi-rs/tree/main/demo

closing this now.

PR #137, opened 2024-09-12
test: obtain fixture artifact path programmatically

prabodh1194:
## Description

I am setting up my dev environment in RustRover IDE and wanted to run the pytest from the IDE's command runner. However, the default setup was giving me some issue related to the `.zip` table artifact. To work around this, I have colocated a python script with the artifact so that I can deterministically get the correct path of the corresponding zip file. 

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 # python
 .env
 venv
+.venv
 **/.python-version
 __pycache__
 
--- a/python/tests/table/table_path.py
+++ b/python/tests/table/table_path.py
@@ -16,27 +16,9 @@
 #  under the License.
 
 import os
-import zipfile
-from pathlib import Path, PosixPath
-
-import pytest
-from _pytest.fixtures import SubRequest
-
-from tests.table.table_path import get_table_path
+from pathlib import Path
 
 
-def _extract_testing_table(zip_file_path: Path, target_path: PosixPath) -> str:
-    with zipfile.ZipFile(zip_file_path, "r") as zip_ref:
-        zip_ref.extractall(target_path)
-    return os.path.join(target_path, "trips_table")
-
-
-@pytest.fixture(
-    params=[
-        "0.x_cow_partitioned",
-    ]
-)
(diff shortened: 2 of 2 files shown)


prabodh1194:
@KnightChess should I go ahead & add a pre-commit hook for `ruff format` ?

KnightChess (commented):
@prabodh1194 I think the ci already has a code check

prabodh1194:
fair enough. thank you.

python/tests/conftest.py
```diff
@@ -17,12 +17,15 @@
...
+from pathlib import Path, PosixPath
 
 import pytest
+from _pytest.fixtures import SubRequest
 
+from tests.table.table_path import get_table_path
 
-def _extract_testing_table(zip_file_path, target_path) -> str:
+
+def _extract_testing_table(zip_file_path: Path, target_path: PosixPath) -> str:
```
xushiyan, line 28: PosixPath restricted this from running on windows? we'll need to support test and build on windows too

xushiyan (commented):
Thanks for raising this. 1 comment to address on the OS compatibility.

Building on top of this, we actually want to use `crates/tests` the centralized place for all testing tables. We already have enums for providing the pre-generated table there. In python codebase, we just need to use pyo3 to bind the crate over so we don't need to maintain separate testing data and module in python.

xushiyan:
@prabodh1194 I created https://github.com/apache/hudi-rs/issues/141 as an extended work for this. If you're interested, feel free to pick it up.

python/tests/conftest.py
prabodh1194: Oh. Got it. Lemme figure out how to test this

prabodh1194:
thanks. lemme try.

xushiyan:
> @prabodh1194 I created #141 as an extended work for this. If you're interested, feel free to pick it up.

We will cover this PR's intent in #141 so we can close this

Issue #136, opened 2024-09-10
PR title linting does not support scope


xushiyan:
Like in this PR

https://github.com/apache/hudi-rs/actions/runs/10780177399/job/29895341649?pr=135

it failed because `build(deps): ...` format was not supported.

We should support this to allow specifying scope of the change.

PR #138, opened 2024-09-14
chore(ci): fix PR title linting to support change scope

kazdy:
## Description
`line 3: syntax error near unexpected token '('`

The syntax error near the unexpected token '(') is due to the parentheses being treated as part of the shell's syntax. Enclosing the entire string in quotes should fix it.

fixes: #136

- [x] N/A
- [x] Manual tests


xushiyan (approved):
Thanks for the quick fix!

PR #149, opened 2024-09-21
build(deps): upgrade datafusion from 0.41 to 0.42

### LINKED ISSUE #143: Upgrade arrow-rs, object_store, datafusion, pyo3 versions
Upgrade the dependencies, and resolve any build issue.

kazdy:
## Description

Upgrade datafusion from 0.41 to 0.42 and its dependencies (arrow, pyo3).
gil-refs are deprecated in pyo3 0.22.2, therefore I removed it, it was causing arrow compilation to fail.
Drop support for Python3.8 as it won't receive security updates starting next month.
Bump required python to min. 3.9.

resolves #143

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -34,7 +34,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pyarrow>=17",
+    "pyarrow>=11.0.0",
     "pyarrow-hotfix",
 ]
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -23,19 +23,18 @@ build-backend = "maturin"
 name = "hudi"
 description = "Native Python binding for Apache Hudi, based on hudi-rs."
 urls = { repository = "https://github.com/apache/hudi-rs/tree/main/python/" }
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 keywords = ["apachehudi", "hudi", "datalake", "arrow"]
 license = "Apache License 2.0"
 classifiers = [
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pyarrow>=8",
+    "pyarrow>=17",
     "pyarrow-hotfix",
 ]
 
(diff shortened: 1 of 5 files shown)


kazdy:
Datafusion plan changed a bit:
image

python/pyproject.toml
```diff
@@ -23,19 +23,18 @@ build-backend = "maturin"
...
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pyarrow>=8",
+    "pyarrow>=17",
     "pyarrow-hotfix",
```
xushiyan, line 38: this is the minimum required version for pyarrow, isn't making it at least 17 too aggressive? also can you check if pyarrow-hotfix still needed if we upgrade to a newer version here

xushiyan (commented):
Sounds good to upgrade python version too. having 1 question on pyarrow

python/pyproject.toml
kazdy: pyarrow-hotfix is not needed since 0.14.1

I don't have a strong opinion on upgrading this, it seemed like a good idea, but I can revert this

xushiyan: thanks for the useful info. what i meant was: this would make hudi-rs pyarrow requirement too strict and exclude people to use when they have lower pyarrow version. I'd align with datafusion's version requirement since we have the integration and it's not too old or too new

python/pyproject.toml
```diff
@@ -23,19 +23,18 @@ build-backend = "maturin"
...
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pyarrow>=8",
+    "pyarrow>=17",
```
xushiyan, line 37:
```suggestion
    "pyarrow>=11.0.0",
```

xushiyan, line 37:
```suggestion
    "pyarrow>=11.0.0",
```

python/pyproject.toml
kazdy: sure, sounds good thanks

xushiyan (approved):

python/pyproject.toml
```diff
@@ -23,19 +23,18 @@ build-backend = "maturin"
...
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pyarrow>=8",
+    "pyarrow>=11.0.0",
```
xushiyan, line 37: we need to update in-code version check also, in `test_table_read.py`

Issue #150, opened 2024-09-24
Create `HudiTableFactory` implementing DataFusions `TableProviderFactory`


matthewmturner:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

This isnt a bug report - its a feature request but i didnt see a way to submit a feature request.

I would like to be able to register hudi tables with datafusion like so:

```
CREATE EXTERNAL TABLE my_table STORED AS HUDITABLE LOCATION '/path/to/table';
```

If hudi-rs provided a `TableProviderFactory` then we could register and use that (This is how i [currently](https://github.com/datafusion-contrib/datafusion-dft/blob/main/src/extensions/deltalake.rs) register deltalake tables and I would like to do something similar for hudi).

### Steps To Reproduce

Not a bug

### Expected behavior

I can register tables to datafusion like so:

```
CREATE EXTERNAL TABLE my_table STORED AS HUDITABLE LOCATION '/path/to/table';
```

### Software information

N/A

xushiyan:
@matthewmturner thanks for raising this! would you be able to review the linked PR by @kazdy please?

matthewmturner:
@xushiyan sure, checking it out

matthewmturner:
Looks good but when i try [integrating](https://github.com/datafusion-contrib/datafusion-dft/pull/189) in my app i get the below error (which isnt surprising since we arent on the same version of datafusion).  Of course thats not a blocker for this though :). Excited to see this though

<img width="1302" alt="image" src="https://github.com/user-attachments/assets/bbb9c805-6637-441a-acbf-17d7e694520e">

xushiyan:
@matthewmturner we just landed this feature in main. can you pls test it out with your PR? note that we choose `HUDI` as the name to stay consistent. See https://github.com/apache/hudi-rs/pull/162#discussion_r1799914280. thanks.

matthewmturner:
@xushiyan sure will do, do you have any recommendations for what test data to use?  i just need something simple to test a basic query

xushiyan:
@matthewmturner yes we keep a bunch of zipped test tables in https://github.com/apache/hudi-rs/tree/main/crates/tests/data/tables

you can unzip and use them directly.

matthewmturner:
Okay, going to work on this tonight / tomorrow.  I may ping you for a review if you dont mind.

matthewmturner:
[...]

PR #162, opened 2024-10-11
feat: implement datafusion TableProviderFactory

kazdy:
## Description

resolves: #150 

This allows datafusion users to register an external hudi table and query it like so:

```sql
CREATE EXTERNAL TABLE trips STORED AS HUDI LOCATION 'path';

SELECT * FROM trips;
```

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/crates/datafusion/src/lib.rs
+++ b/crates/datafusion/src/lib.rs
@@ -179,9 +179,6 @@ mod tests {
     use std::path::Path;
     use std::sync::Arc;
 
-    // use datafusion::catalog::TableProviderFactory;
-    // use datafusion::execution::context::SessionState;
-    // use datafusion::execution::runtime_env::RuntimeEnv;
     use datafusion::execution::session_state::SessionStateBuilder;
     use datafusion::prelude::{SessionConfig, SessionContext};
     use datafusion_common::ScalarValue;
--- a/crates/datafusion/src/lib.rs
+++ b/crates/datafusion/src/lib.rs
@@ -373,4 +373,37 @@ mod tests {
             verify_data(&ctx, &sql, test_table.as_ref()).await
         }
     }
+
+    #[tokio::test]
+    async fn datafusion_read_external_hudi_table_with_replacecommits() {
+        for (test_table, planned_input_partitions) in
+            &[(V6SimplekeygenNonhivestyleOverwritetable, 1)]
+        {
+            println!(">>> testing for {}", test_table.as_ref());
+            let ctx = prepare_session_context_with_table_factory().await;
+            let base_path = test_table.path();
+
+            let create_table_sql = format!(
+                "CREATE EXTERNAL TABLE {} STORED AS HUDITABLE LOCATION '{}' OPTIONS ('hoodie.read.input.partitions' '2');",
+                test_table.as_ref(),
+                base_path
+            );
+
+            let _ = ctx
+                .sql(create_table_sql.as_str())
+                .await
+                .expect("Failed to register table");
+
+            let sql = format!(
+                r#"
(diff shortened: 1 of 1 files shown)


matthewmturner (approved):

kazdy:
@xushiyan can I get another CI run, I added a test for when options are passed when creating external table, should make codecov happy

crates/datafusion/src/lib.rs
```diff
@@ -287,4 +330,73 @@ mod tests {
...
+            V6SimplekeygenNonhivestyle,
+            V6SimplekeygenHivestyleNoMetafields,
+            V6TimebasedkeygenNonhivestyle,
+        ] {
+            println!(">>> testing for {}", test_table.as_ref());
+            let ctx = prepare_session_context_with_table_factory().await;
+            let base_path = test_table.path();
+
+            let create_table_sql = format!(
+                "CREATE EXTERNAL TABLE {} STORED AS HUDITABLE LOCATION '{}'",
```
xushiyan, line 348: i think `HUDITABLE` is a bit redundant - we already know it's creating a table. More importantly, in other sql engine like spark/flink/hive, we all use `HUDI` to indicate the format. So let's be consistent and simplify this as `HUDI`

crates/datafusion/src/lib.rs
```diff
@@ -287,4 +330,73 @@ mod tests {
...
+            verify_data(&ctx, &sql, test_table.as_ref()).await
+        }
+    }
+
+    #[tokio::test]
+    async fn datafusion_read_external_hudi_table_with_replacecommits() {
+        for (test_table, planned_input_partitions) in
+            &[(V6SimplekeygenNonhivestyleOverwritetable, 1)]
+        {
+            println!(">>> testing for {}", test_table.as_ref());
```
xushiyan, line 375: avoid print statement; tests don't say anything unless they fail

xushiyan (changes requested):

crates/datafusion/src/lib.rs
```diff
@@ -143,12 +143,36 @@ impl TableProvider for HudiDataSource {
     }
 }
 
+pub struct HudiTableFactory;
+
+#[async_trait]
+impl TableProviderFactory for HudiTableFactory {
+    async fn create(
+        &self,
+        _state: &dyn Session,
```
xushiyan, line 152: should also take configs from session?

xushiyan:
@kazdy i'll push quick fixes and then freeze the code for release

crates/datafusion/src/lib.rs
kazdy: other tests in this file do have the same, I just copied that over, should I remove other print statements as well?

crates/datafusion/src/lib.rs
kazdy: I don't see this being used in other implementations (PostgreSQL and duckdb):
https://github.com/datafusion-contrib/datafusion-table-providers/blob/ce93f2c916fcd57d1d5b7339e64963c3a72d3673/src/postgres.rs#L199

https://github.com/datafusion-contrib/datafusion-table-providers/blob/ce93f2c916fcd57d1d5b7339e64963c3a72d3673/src/duckdb.rs#L208

passing it here feels redundant as TableProvider::scan already provides a session, but I'm not sure really, this seem to work as is

crates/datafusion/src/lib.rs
xushiyan: right, this statement is intended to compensate not using parameterized test and give additional info about which one failed. so we can keep this one.

xushiyan (approved):

Issue #139, opened 2024-09-14
feat: support storage_options param when reading from table


kazdy:
To integrate hudi-rs with AWS SDK for Pandas (aws wrangler), we must be able to pass boto_session related aws authentication params (mostly AWS_* params) directly and not only rely on env variable inference.

I want to propose adding an option to handle this: 
```rust
storage_options = {"AWS_ACCESS_KEY_ID": "xxxx", "AWS_SECRET_ACCESS_KEY":"xxxx", "AWS_SECRET_ACCESS_TOKEN":"xxxx"}
hudi_table = HudiTable("/tmp/trips_table", storage_options=storage_options)
records = hudi_table.read_snapshot()
```

Although I want to add this for S3, it should work for other storage backends.
I'm happy to contribute and add this.

xushiyan:
@kazdy sounds good. feel free to take this up and send a pr

kazdy:
I'll wait until #72 gets merged. 
I did the first strawman impl and it requires some refactoring in the Table itself.

@xushiyan I also have some questions about this, maybe you can give me your opinion on these:
1. Should we rename Table to HudiTable?
2. I don't know why Timeline and FileSystemView both use separate storage instances, can't they share it, maybe there's a reason why it's done this way I can't see atm?
3. Does it make sense to introduce something that will hold both Timeline and FileSystemView (basically table state) and expose coherent API?

thanks

xushiyan:
hey @kazdy 

1) we keep name `Table` within `hudi-core` to avoid redundant prefix; everything in `hudi-core` is about Hudi. When import to other crates, we can give it an alias like `HudiTable`. We can also add an alias in `hudi` crate for external facing API when needed. As of now, no strong need for this.

2) Timeline is responsible for data stored in timeline files under `.hoodie/`, and FileSystemView is responsible for the data stored under the table excluding `.hoodie/`. It's good to keep things less coupled, unless there is a need for sharing - it's a stateless client performing IO anyway. Maybe you can make a case about why sharing it?

3) Currently Table holds Timeline and FileSystemView. You want to elaborate on what you meant by coherent API?

PR #163, opened 2024-10-11
feat: add `TableBuilder` API for passing options

kazdy:
## Description

Add `TableBuilder` to handle Table construction.

Allow to explicitly set hudi_options and storage_options via builder in rust or via python bindings when creating new Table instance.

```rust
let hudi_table = TableBuilder::from_base_uri("/tmp/hudi_table")
  .with_hudi_options([("hoodie.read.as.of.timestamp", "0")])
  .with_storage_options([("AWS_REGION", "us-east-1")])
  .build()
```

similarly for python

```python
hudi_table = HudiTableBuilder.from_base_uri(
  "/tmp/hudi_table"
  ).with_hudi_options({
    "hoodie.read.as.of.timestamp": "0"
  }).with_storage_options({
    "AWS_REGION": "us-east-1"
  }).build()
```

resolves: #139 

- [ ] N/A
- [ ] Manual tests


kazdy:
@xushiyan please let me know if this is a direction you're fine with

crates/core/src/table/mod.rs
```diff
@@ -180,91 +169,6 @@ impl Table {
             .register_object_store(runtime_env.clone());
     }
 
-    async fn load_configs<I, K, V>(
```
xushiyan, line 183: some of these util functions were updated in main

crates/core/src/table/builder.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+    pub fn with_hudi_options(mut self, hudi_options: HashMap<String, String>) -> Self {
+        self.hudi_options = Some(hudi_options);
+        self
+    }
+
+    pub fn with_storage_options(mut self, storage_options: HashMap<String, String>) -> Self {
+        self.storage_options = Some(storage_options);
+        self
+    }
```
xushiyan, line 57: a builder pattern is helpful for users. can you design such that it takes in generic options? we don't have to distinguish hudi option vs storage option from the user side, the build() function should handle it internally.

xushiyan (commented):
there are some refactoring around this. can you check out https://github.com/apache/hudi-rs/pull/161 first, and adjust this PR accordingly please?

python/hudi/_internal.pyi
```diff
@@ -38,7 +38,8 @@ class HudiTable:
     def __init__(
         self,
         table_uri: str,
-        options: Optional[Dict[str, str]] = None,
+        hudi_options: Optional[Dict[str, str]] = None,
+        storage_options: Optional[Dict[str, str]] = None,
```
xushiyan, line 42: you can revert this change - we keep python and rust api as close as possible

python/src/internal.rs
```diff
@@ -101,12 +102,18 @@ pub struct HudiTable {
...
-    fn new(table_uri: &str, options: Option<HashMap<String, String>>) -> PyResult<Self> {
-        let _table = rt().block_on(Table::new_with_options(
-            table_uri,
-            options.unwrap_or_default(),
-        ))?;
+    #[pyo3(signature = (table_uri, hudi_options = None, storage_options = None))]
+    fn new(
+        table_uri: &str,
+        hudi_options: Option<HashMap<String, String>>,
+        storage_options: Option<HashMap<String, String>>,
```
xushiyan, line 109: this can be reverted as well

python/tests/test_table_read.py
```diff
@@ -133,7 +133,9 @@ def test_sample_table(get_sample_table):
         },
     ]
 
-    table = HudiTable(table_path, {"hoodie.read.as.of.timestamp": "20240402123035233"})
+    table = HudiTable(
+        table_path, hudi_options={"hoodie.read.as.of.timestamp": "20240402123035233"}
+    )
```
xushiyan, line 138: ditto

xushiyan:
@kazdy i'm cutting release branch within 1 day. do you think you can update this soon?

crates/core/src/table/builder.rs
kazdy: Ok I'll do it, but I'd like to still give users (and myself as a user as well) option to be specific.
I wanted to have storage_options because from the first look I did not figure out options handle both hudi and storage opts.

So it'll be 
TableBuilder::with_options() -> handle generic options
TableBuilder::with_hudi_options()
TableBuilder::with_storage_options()

I hope that's fine

kazdy:
> @kazdy i'm cutting release branch within 1 day. do you think you can update this soon?

yes, I'll give it a try

python/hudi/_internal.pyi
kazdy: will do, I'll introduce HudiTableBuilder in python bindings instead

crates/core/src/table/builder.rs
xushiyan: @kazdy can you please clarify the benefit of using 

```
TableBuilder::with_hudi_options()
TableBuilder::with_storage_options()
```

when there is a generic api to take options and resolve internally, why would you still want to use an explicit api? using a generic api is one less thing caller needs to worry about. From API design pov, it'll be redundant unless being explicit gives additional logic behind it.

xushiyan: if you add additional checks, like for hudi_options, do some validation (there is a util function for this) for the input, and for storage options, check they have certain prefixes, then that can be helpful in preventing mistakes

xushiyan: but still, builder pattern usually does validation in the end when call build(), we don't want to panic during the building phase of with_xxx(). One point i can think of is, making input taken from explicit APIs have higher precedence than those taken by generic ones. This should be handled in build()

kazdy: I think TableBuilder can give more control to the user and be more specific. I'm fine with having a generic API for ease of use, but I still think having this explicit distinction between hudi and storage options is nice and easy to understand when using the API. I was a bit confused at first and wondered if/where can I pass storage options.

Still Table has ::new_with_options() which is generic so that TableBuilder can be more specific?

I agree about validation being done in ::build().
I'm not sure if I'll be able to add more validation now, but I can give it a try in the next PR?

xushiyan: > I'm not sure if I'll be able to add more validation now, but I can give it a try in the next PR?

yup sounds good

xushiyan:
@kazdy I'll push some fixes and then get this landed for release

python/tests/test_table_builder.py
kazdy, line 1: @xushiyan please take a look

python/tests/test_table_read.py
kazdy, line 1: @xushiyan please take a look, I took the freedom to split this one big test to a few smaller ones, I hope that's fine with you

crates/core/src/table/builder.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    async fn imbue_table_properties(
+        options: &mut HashMap<String, String>,
+        storage: Arc<Storage>,
+    ) -> anyhow::Result<()> {
+        let bytes = storage.get_file_data(".hoodie/hoodie.properties").await?;
+        let table_properties = parse_data_for_options(&bytes, "=")?;
+
+        // TODO: handle the case where the same key is present in both table properties and options
```
kazdy, line 161: in such case, which option should take precedence?
@xushiyan

python/tests/test_table_read.py
xushiyan: yup good improvement

crates/core/src/table/builder.rs
xushiyan: it's a bit complicated, there are difference cases pertaining to individual config - i'll fix this todo

xushiyan (approved):

Issue #165, opened 2024-10-12
Partition filter against string or date field should  strip `'` around the value


xushiyan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of the bug

Partition filter currently works for `foo > A` for string field. To match sql syntax, it should be `foo > 'A'`.  But current logic won't trim the single quotes and take `'A'` as a whole for the value. Similarly for date types, it should parse `date = '2020-01-01'` correctly.

### Steps To Reproduce

Run with partition filters like `foo > A` for a string field, and `date = '2020-01-01'` for a date field.

### Expected behavior

Single quotes should be stripped for applicable field types.

### Software information

- Operating system:
- Rust version:
- Project version:

xushiyan:
hey @KnightChess do you happen to have some time to work on this issue? hope to get this in 0.2.0

KnightChess:
@xushiyan hi, sorry, I've been quite busy these past few weeks and haven't had time to follow up on this issue.

xushiyan:
@KnightChess no problem

kazdy:
@xushiyan please take a look at #170 I took an attempt to fix this

PR #170, opened 2024-10-13
fix: simplify partition filter format by taking tuple of strings

kazdy:
## Description

fixes: #165

- [ ] N/A
- [ ] Manual tests


crates/core/src/table/partition.rs
```diff
@@ -265,13 +265,36 @@ impl PartitionFilter {
             safe: false,
             format_options: Default::default(),
         };
+
+        let value = match data_type {
+            DataType::Date32 => Self::trim_single_quotes(value),
+            DataType::Date64 => Self::trim_single_quotes(value),
+            DataType::Utf8 => Self::trim_single_quotes(value),
+            DataType::LargeUtf8 => Self::trim_single_quotes(value),
+            DataType::Utf8View => Self::trim_single_quotes(value),
```
kazdy, line 274: time and timestamp types should be included too?

xushiyan: yes it should also applies to any temporal type that is supposed to be parsed from string in sql.

crates/core/src/table/partition.rs
```diff
@@ -265,13 +265,34 @@ impl PartitionFilter {
...
             &cast_options,
         )?))
     }
+
+    fn trim_single_quotes<'a>(s: &'a [&'a str; 1]) -> [&'a str; 1] {
+        let trimmed = s[0]
+            .strip_prefix('\'')
+            .unwrap_or(s[0])
+            .strip_suffix('\'')
+            .unwrap_or(s[0]);
```
xushiyan, line 292: this does not complain when value is not wrapped with `'`. like it's accepting `foo'`, `'foo`, or `foo`. can you follow and check what would be valid in like spark sql and datafusion?

crates/core/src/table/partition.rs
kazdy: Alright, I'll add that later today

crates/core/src/table/partition.rs
```diff
@@ -506,4 +525,75 @@ mod tests {
...
+    fn test_do_not_strip_one_side_single_quoted_value_for_filter() {
+        let data_type = DataType::Utf8;
+        let schema = Schema::new(vec![Field::new("category", data_type.clone(), false)]);
+
+        let cast_options = CastOptions {
+            safe: false,
+            format_options: Default::default(),
+        };
+
+        let value = StringArray::from(vec!["'foo"]);
```
xushiyan, line 587: what i meant in the previous comment was: shouldn't this fail due to invalid sql syntax? can you check how these work with other sql engines? for e.g., how would datafusion treat this when query parquet table?

kazdy: im not sure I follow
how the query would look like?

For sure this will work properly:
`SELECT * FROM table where part='2022-01-01'`
but
`SELECT * FROM table where part='2022-01-01`
will fail when parsing the query.

kazdy: as for strings this should result in an error?
`SELECT * FROM table where part=A`
and this should pass:
`SELECT * FROM table where part=A`?

kazdy: i'm not sure about this, shouldn't this be the responsibility of a query engine to parse this?
The partition pruner could be implemented specifically for a query engine first?

kazdy: @xushiyan what if we flip it a bit and make 
PartitionFilter::try_from() take a 3 element tuple containing key, operator, value instead?
Then we can (based on the schema) simply assume the type and don't need worry about the single quotes.

so:
`PartitionFilter::try_from(("part", "=", "A"))`
`PartitionFilter::try_from(("part", "=", "2022-01-01"))`
`PartitionFilter::try_from(("part", "IN", ["2022-01-01", "2023-02-02"]))`

Then we simplify and leave the parsing to the user or system that uses that, but it's simple enough not to be frustrating to use.

kazdy: @xushiyan please take a look and let me know what you think.
Imo this simplifies a lot on hudi-rs side, still pretty ok from the user's point of view.

Now the other way around it should be easy to generate these filter strings if required by some engine etc.

xushiyan: yup the approach looks good. it's good to start with something simple, then iterate on it. getting back to this. will review later

xushiyan (approved):

python/src/internal.rs
```diff
@@ -30,12 +30,17 @@ use hudi::file_group::FileSlice;
...
+        .map(|(a, b, c)| (a.as_str(), b.as_str(), c.as_str()))
+        .collect()
+}
+
+macro_rules! vec_to_slice {
     ($vec:expr) => {
-        &$vec.iter().map(AsRef::as_ref).collect::<Vec<_>>()
+        &convert_vec_to_slice(&$vec)[..]
     };
 }
```
xushiyan, line 43: while you're here, can you move this util to a common module in hudi-core, with a basic UT covering it?

kazdy: I see that some modules in hudi core have utils.rs, I followed the same for this python crate for now

xushiyan: @kazdy this is a generic util that should be placed under common for all modules to use. python does not mimic rust for every util, it imports when needed.

kazdy: ok, done

PR #203, opened 2024-11-29
feat: support filter pushdown for datafusion

### LINKED ISSUE #160: Integrate with datafusion to support filters pushdown from SQL
Follow-up feature of #47 

Integrate with datafusion on the predicate expression parsing and support pushdown predicates from SQL.

jonathanc-n:
## Description

Added `Expr` to `PartitionFilter` conversion to pass in filters. Datafusion will pass down all filters for now using `supports_filters_pushdown` and will filter after partition filters.

Closes #160.

## Steps forward
I noticed the pr was getting a bit big to be reviewed all at once, here are some things that will be worked on afterwards:
- Support more operators
- Support more Datafusion Expressions
- Enhance the python implementation for filter conversion
- Make easier to create `PartitionFilters` for testing, etc.

Added unit tests for all added functionality


crates/datafusion/src/utils/exprs_to_filter.rs
```diff
@@ -0,0 +1,346 @@
...
+                continue;
+            }
+        }
+    }
+
+    partition_filters
+}
+
+/// Converts a binary expression (`Expr::BinaryExpr`) into a `PartitionFilter`.
+fn convert_binary_expr(
```
xushiyan, line 61: instead of helper functions, can we implement `TryFrom` for Hudi's generic filter struct to convert Expr to Hudi filter?

python/src/internal.rs
```diff
@@ -163,10 +164,13 @@ impl HudiTable {
         filters: Option<Vec<(String, String, String)>>,
```
xushiyan, line 164: we should keep python API aligned with rust API. if we introduce a generic filter struct, we would need to bind it over here using pyo3, and then use it in the API

crates/core/src/table/partition.rs
```diff
@@ -40,14 +39,11 @@ pub struct PartitionPruner {
...
-        and_filters: &[(&str, &str, &str)],
+        and_filters: &[PartitionFilter],
         partition_schema: &Schema,
         hudi_configs: &HudiConfigs,
     ) -> Result<Self> {
-        let and_filters = and_filters
-            .iter()
-            .map(|filter| PartitionFilter::try_from((*filter, partition_schema)))
-            .collect::<Result<Vec<PartitionFilter>>>()?;
+        let and_filters = and_filters.to_vec();
```
xushiyan, line 46: partition pruner should be one to decide which filter is a valid "partition filter" by checking against the partition schema. Other filters will be ignored here (but will be used by storage layer push down logic which we have not implemented yet). so to summarize the high-level flow is like:
- Hudi filters are constructed from datafusion exprs and pass to the table APIs
- PartitionPruner is then constructed, which could be empty in case no filter is applied to partition fields

crates/datafusion/src/lib.rs
```diff
@@ -176,6 +215,22 @@ impl TableProvider for HudiDataSource {
...
+
+    fn supports_filters_pushdown(
+        &self,
+        filters: &[&Expr],
+    ) -> Result<Vec<TableProviderFilterPushDown>> {
+        filters
+            .iter()
+            .map(|expr| {
+                if self.can_push_down(expr) {
+                    Ok(TableProviderFilterPushDown::Exact)
```
xushiyan, line 227: since we only do partition pruning, we can't guarantee all filters used, so it should be `Inexact`

python/src/internal.rs
jonathanc-n: Should this be done in another issue? (introducing the generic struct. one of your reviews above mention as well)

crates/core/src/table/mod.rs
```diff
@@ -194,7 +195,7 @@ impl Table {
     pub async fn get_file_slices_splits(
         &self,
         n: usize,
-        filters: &[(&str, &str, &str)],
+        filters: &[PartitionFilter],
```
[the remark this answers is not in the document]
jonathanc-n: I made the fixes but I was a little confused how to deprecate this. Should I create another similar function?

jonathanc-n:
fixed conflicts

crates/core/src/table/mod.rs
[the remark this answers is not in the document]
xushiyan: usually just mark a function`#[deprecated]`, and create the replacing function

xushiyan:
> @xushiyan Should we add some docs to see how to run codecov locally? It takes a lot of time out of the review process

Yup having some command in `Makefile` to general html report locally would be beneficial. Codecov aggregates both rust and python, for local mode, not sure how easy to configure that, may start with separate report.

xushiyan:
I'm going to take a pass by today and fix wherever necessary, and then merge it.

xushiyan (approved):

PR #238, opened 2025-01-08
feat: improve partition pruning during file listing

### LINKED ISSUE #205: Improve partition pruning during file listing
During partition loading (https://github.com/apache/hudi-rs/blob/d90abed1a8d7da54bbde81aad473d4a923b5c597/crates/core/src/table/fs_view.rs#L60) , the file listing can be shortcut by partition pruner to reduce potentially a lot of file listing calls.

TheR1sing3un:
## Description

1. improve partition pruning during file listing
2. fix incorrect adding `FileSystemView::partition_to_file_groups`

Closes #205

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/crates/core/src/table/fs_view.rs
+++ b/crates/core/src/table/fs_view.rs
@@ -17,13 +17,14 @@
  * under the License.
  */
 
-use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
-
 use crate::config::HudiConfigs;
 use crate::file_group::base_file::BaseFile;
 use crate::file_group::{FileGroup, FileSlice};
 use crate::storage::{get_leaf_dirs, Storage};
+use async_recursion::async_recursion;
+use std::collections::{HashMap, HashSet};
+use std::path::PathBuf;
+use std::sync::Arc;
 
 use crate::config::read::HudiReadConfig::ListingParallelism;
 use crate::error::CoreError;
@@ -32,6 +33,8 @@ use crate::table::partition::PartitionPruner;
 use crate::Result;
 use dashmap::DashMap;
 use futures::stream::{self, StreamExt, TryStreamExt};
+
+pub const HOODIE_PARTITION_METAFILE_NAME: &str = ".hoodie_partition_metadata";
 
 /// A view of the Hudi table's data files (files stored outside the `.hoodie/` directory) in the file system. It provides APIs to load and
 /// access the file groups and file slices.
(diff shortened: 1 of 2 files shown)


TheR1sing3un:
@xushiyan Hi, thank you for reviewing it for me when you have time!

xushiyan:
thanks @TheR1sing3un will go through this over the weekend.

xushiyan:
I'm looking into the conflicts as I have more context

crates/core/src/table/fs_view.rs
```diff
@@ -73,34 +77,84 @@ impl FileSystemView {
...
+        Ok(partition_paths.into_iter().collect())
+    }
+
+    #[async_recursion]
+    async fn list_partition_paths_with_leveled_pruning(
+        storage: &Storage,
+        path: String,
+        partition_pruner: &PartitionPruner,
+        current_level: usize,
+    ) -> Result<Vec<String>> {
```
xushiyan, line 104: can this be merged with the `get_leaf_dirs()`? so that we don't need to go 2 paths that one does multi-level pruning and the other does not. We can by default use multi-level pruning and early return when possible.

Check out this line https://github.com/apache/hudi-rs/pull/251/files#diff-a6c5a16b1d83ac9b5d186863a123fd1b273daf8710c12444c9f1b6f6fec96914R140

we probably need to move `get_leaf_dirs()` into `FileLister` such that it's the lister's responsibility to use pruner to check partition segments.

PartitionPruner owns the partition schema, so we should make use of it to understand which level is currently parsed and how many levels there should be. 

To make code logic easy to understand, maybe iterative is better than recursive here - just loop through all the top level dirs and go down according to the schema.

xushiyan (commented):
@TheR1sing3un thanks for the good work. i cleaned up relevant code paths in https://github.com/apache/hudi-rs/pull/251/ which the improvement here can be built upon. Left concrete ideas in the comments below.

xushiyan:
Hi @TheR1sing3un, thanks again for the work here and apologies for the long delay.

After re-evaluating, I'm going to close this PR. Summary of the analysis:

**Code path scope.** This optimization only runs in the no-MDT file-listing path. With the metadata table FILES partition enabled (the production recommendation), partition discovery comes from metadata records and bypasses filesystem listing entirely.

**Architecture has shifted.** Since this PR was opened, the listing logic moved out of `FileSystemView` into a dedicated `FileLister` (`crates/core/src/table/listing.rs`), so the patch would need a substantial rework to apply.

**ROI of the leveled-pruning approach as proposed:**
- Gain: ~10–20x fewer sequential LIST calls only when (no MDT) AND (remote object store) AND (selective partition filter). Zero gain otherwise.
- Cost: ~150 LOC of production code, a new `should_include_with_level` / `parse_segment_with_level` API surface that has to stay in sync with `should_include` forever, breaking `PartitionPruner::schema` encapsulation, and changing `list_file_groups_for_partition` to return a tuple.

**Cheaper alternatives** that capture the same wins (and more) with much less surface area:
1. Parallelize `get_leaf_dirs` in `crates/core/src/storage/mod.rs` (the recursive descent is currently sequential). This benefits every no-MDT query, not just the filtered ones.
2. Add an optional per-level predicate to `get_leaf_dirs` so pruning happens during descent — same effect as the leveled approach without a parallel pruner API.

I've re-scoped #205 along these lines and left it open. Would love to have you pick it up again, or anyone else in the community.

Closing this PR. Thanks for the contribution and for raising the perf concern in the first place — the underlying issue stays on the roadmap.

TheR1sing3un:
> Closing this PR. Thanks for the contribution and for raising the perf concern in the first place — the underlying issue stays on the roadmap.

Great! Thanks for your review ~

PR #302, opened 2025-02-24
feat: iso8601 support in Time-travel & Incremental API

### LINKED ISSUE #261: Incremental API to support more timestamp formats
### Description of the improvement

Currently `Table::read_incremental_records()` expects start and end timestamps to be in the Hudi timeline format, i.e., `yyyyMMddHHmmsssss` or the legacy `yyyyMMddHHmmss`. We should support parsing more timestamp formats.

### Expected behavior

- To support parsing strings in the form of epoch time (including seconds, milliseconds, microseconds, nanoseconds) and convert to Hudi timeline format for further processing.

- To support parsing strings in the form of ISO8601 format like
  - `2019-01-23T12:34:56.123456789+00:00`
  - `2019-01-23T12:34:56.123456+00:00`
  - `2019-01-23T12:34:56.123+00:00`
  - `2019-01-23T12:34:56+00:00`
  - `2019-01-23T12:34:56Z`
  - `2019-01-23`

### Additional context

The conversion to Hudi timeline format should consider the hudi table's timeline timezone config.

hanbings:
## Description

Added support for ISO8601 time format as described in issues, and overwrote other `*as_of()` methods in the same module.

closes #261

- [ ] N/A
- [x] Manual tests

### DIFF
--- a/crates/core/src/table/mod.rs
+++ b/crates/core/src/table/mod.rs
@@ -241,8 +241,9 @@ impl Table {
         timestamp: &str,
         filters: &[(&str, &str, &str)],
     ) -> Result<Vec<Vec<FileSlice>>> {
+        let timestamp = format_timestamp(timestamp);
         let filters = from_str_tuples(filters)?;
-        self.get_file_slices_splits_internal(n, timestamp, &filters)
+        self.get_file_slices_splits_internal(n, &timestamp, &filters)
             .await
     }
 
@@ -295,8 +296,9 @@ impl Table {
         timestamp: &str,
         filters: &[(&str, &str, &str)],
     ) -> Result<Vec<FileSlice>> {
+        let timestamp = format_timestamp(timestamp);
         let filters = from_str_tuples(filters)?;
-        self.get_file_slices_internal(timestamp, &filters).await
+        self.get_file_slices_internal(&timestamp, &filters).await
     }
 
     async fn get_file_slices_internal(
@@ -406,8 +408,9 @@ impl Table {
         timestamp: &str,
         filters: &[(&str, &str, &str)],
     ) -> Result<Vec<RecordBatch>> {
+        let timestamp = format_timestamp(timestamp);
         let filters = from_str_tuples(filters)?;
-        self.read_snapshot_internal(timestamp, &filters).await
+        self.read_snapshot_internal(&timestamp, &filters).await
     }
 
     async fn read_snapshot_internal(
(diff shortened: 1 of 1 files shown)


hanbings:
It seems that there are some dependency conflicts here (in Github Actions). 
How should we proceed with the next step?

xushiyan:
hi @hanbings sorry for the delay - the issue should have been resolved, could you rebase master please?

crates/core/src/table/mod.rs
```diff
@@ -474,6 +481,36 @@ impl Table {
...
+/// The function attempts to parse the input timestamp string (`timestamp`) into a `DateTime` object.
+/// It first tries parsing the ISO 8601 format, and if that fails, it tries two other formats: `yyyyMMddHHmmSSSSS`
+/// and `yyyyMMddHHmmSS`. If none of the formats match, it returns the original input string.
+///
+/// # Arguments
+/// - `timestamp`: The input timestamp string to be formatted.
+///
+/// # Returns
+/// A string formatted as `yyyyMMddHHmmSSSSS`. If the input cannot be parsed, the original string is returned.
+fn format_timestamp(timestamp: &str) -> String {
```
xushiyan, line 495: i'd expect this returning a Result - if users pass a timestamp str that can't be converted to hudi timeline format properly, then the flow should early return.

The user-provided timestamp should also conform to timezone config of the table - this is by default utc; but it could be configured as `local`.

crates/core/src/table/mod.rs
```diff
@@ -474,6 +481,36 @@ impl Table {
...
+    }
+
+    let formats = ["yyyyMMddHHmmSSSSS", "yyyyMMddHHmmSS"];
+    for format in formats.iter() {
+        if let Ok(datetime) = DateTime::parse_from_str(timestamp, format) {
+            return datetime.format("%Y%m%d%H%M%S%3f").to_string();
+        }
+    }
+
+    if timestamp.len() == 10 && timestamp.chars().all(|c| c.is_digit(10) || c == '-') {
```
xushiyan, line 507: what cases are meant for this check? this checking logic is hard to maintain - please use proper pattern matching to improve the case handling

crates/core/src/table/mod.rs
hanbings: I have added an is_utc parameter to this function in the new commit to mark whether the utc time zone is used, and changed the return value to `Result`, it will return a CoreError::Timeline when parsing errors occur.

crates/core/src/table/mod.rs
hanbings: This check is for short time formats like `2019-01-23`. I haven’t come up with a better way to handle this yet — do you have any suggestions?

xushiyan (approved):

PR #394, opened 2025-07-13
test: add basic v8 test tables

xushiyan:
## Description

Add test tables of v8.

For https://github.com/apache/hudi-rs/issues/379

## How are the changes test-covered

- [x] N/A
- [ ] Automated tests (unit and/or integration tests)
- [ ] Manual tests
  - [ ] Details are described below

PR #395, opened 2025-07-13
feat: add LSM timeline support with timeline layout management for old and new table versions

codope:
## Description

Implements LSM timeline reading capability for the hudi-rs core to support both v6 and v8 timeline formats, following Java implementation patterns with timeline layout abstraction. This is the foundation PR and we will follow up with future work on LSM timeline history reading, support for multiple layers of timeline compaction and manifest parsing for compacted timeline files.

### Core Implementation
- **Timeline Layout Architecture**: Added `TimelineLayoutType` enum to manage different timeline storage formats.
- **Active Timeline**: Implemented `ActiveTimeline` for older table versions with layout version 1 (existing format).
- **LSM Timeline**: Added `LsmTimeline` foundation for newer table version 8 with layout version 2 (LSM format).
- **Factory Pattern**: Created `TimelineLayoutFactory` for creating appropriate timeline implementations based on table configuration.

```rust
Timeline {
    layout: TimelineLayoutType::Active(ActiveTimeline) |
            TimelineLayoutType::Lsm(LsmTimeline)
}
```

- **layout v1**: Uses `ActiveTimeline` (reads from `.hoodie/` directory)
- **layout v2**: Uses `LsmTimeline` (foundation for `.hoodie/timeline/` LSM format)

### Configuration Integration
- **Version Detection**: Automatic timeline layout selection based on `hoodie.table.version` and `hoodie.timeline.layout.version`
- **Graceful Fallback**: Handles invalid table configurations by defaulting to active timeline layout.
- **Backward Compatibility**: Maintains full compatibility with existing v6 table APIs.

Related to #8 

- **250/250 core tests** passing  
- **Integration tests** for table reading, snapshot queries, and file slices
- **Backward compatibility** verified with existing test data


crates/core/src/file_group/reader.rs
```diff
@@ -75,8 +75,8 @@ impl FileGroupReader {
     /// Creates a new reader with the given base URI and options.
     ///
     /// # Arguments
-    ///     * `base_uri` - The base URI of the file group's residing table.
-    ///     * `options` - Additional options for the reader.
+    /// * `base_uri` - The base URI of the file group's residing table.
+    /// * `options` - Additional options for the reader.
```
codope, line 79: Please ignore these changes while reviewing. When I ran `cargo test --workspace` there were quite a few doctests that failed locally. So, I had to fix them.

crates/core/src/timeline/lsm.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn new(_hudi_configs: Arc<HudiConfigs>, storage: Arc<Storage>) -> Self {
+        Self { storage }
+    }
+
+    pub async fn load_instants(
+        &self,
+        selector: &TimelineSelector,
+        desc: bool,
+    ) -> Result<Vec<Instant>> {
+        // For now, fall back to reading from .hoodie/ directory like active timeline
```
codope: Please ignore codecov warnings. We will implement more functionality once #394 lands. For now, we just fallback to v6.

codope (commented):
A few note to reviewers.

cpp/Cargo.toml
```diff
@@ -28,7 +28,7 @@ homepage.workspace = true
 repository = "https://github.com/apache/hudi-rs/tree/main/cpp/"
 
 [lib]
-name = "hudi"
+name = "hudi_cpp"
```
xushiyan: this actually impact the library name which is used in CMakeLists.txt . let's keep unrelated change out from the PR anyway. We try to keep artifact name minimum: we don't need to call it hudi_cpp as a c++ lib, just like we don't need to call the python artifact pyhudi - we just `pip install hudi`

crates/core/src/timeline/active.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::sync::Arc;
+
+/// Active timeline for table versions 5-8.
+/// Reads individual instant files from `.hoodie/` directory.
+#[derive(Debug, Clone)]
+pub struct ActiveTimeline {
+    storage: Arc<Storage>,
+}
+
+impl ActiveTimeline {
```
xushiyan: let's take the opportunity to re-think some of the model design: 

from the callers perspective,  they should not care if the commits should come from active timeline `history/` or `archived/` because these are all timeline internal logic flow. If the caller requests require looking into compacted or archived commits, the timeline instance will need to make a call and fetch the data as needed. So at the interface level, we don't need an explicit `ActiveTimeline`. We can just keep it `Timeline` as the front-facing model.

crates/core/src/timeline/layout.rs
```diff
@@ -0,0 +1,86 @@
...
+use crate::timeline::instant::Instant;
+use crate::timeline::selector::TimelineSelector;
+use crate::Result;
+use std::sync::Arc;
+
+/// Timeline layout implementations for different table versions.
+#[derive(Debug, Clone)]
+pub enum TimelineLayoutType {
+    Active(super::active::ActiveTimeline),
+    Lsm(super::lsm::LsmTimeline),
```
xushiyan: We think of `Timeline` as a top level interface where callers can interact with, the layout is an internal config that resolved based on table version and timeline layout version (as you resolved in `create_layout()`. It'd be a bit odd to see a "Type" struct having a Timeline instance basically for choosing a Timeline to use.

How about we design it like:

- `Timeline` remains as the top-level interface
- When a Timeline is created, it resolves `HudiConfigs`, and create an internal model call `TimelineLoader` based on the configs
- `TimelineLoader` can be an enum, it has these items:

```
LayoutOneActive
LayoutOneArchived
LayoutTwoActive
LayoutTwoCompacted
```

Each loader will implement the logic to load its intended commit files.

- When some APIs are invoked and need to load archived or compacted commits, the Timeline instance will use the needed `Loader`s to do IO
- `Timeline` is responsible for collecting and stitching the `Loader`s' return data and returning final data to callers
- We also avoid calling a LSM timeline since LSM tree is just one impl of Timeline which should be hidden away from Timeline's level

crates/core/src/timeline/lsm.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// LSM timeline for table version 8.
+/// For now, falls back to reading from .hoodie/ directory like active timeline.
+/// TODO: Add LSM timeline history reading from .hoodie/timeline/ directory
+#[derive(Debug, Clone)]
+pub struct LsmTimeline {
+    storage: Arc<Storage>,
+}
+
+impl LsmTimeline {
```
xushiyan: We can keep LSM tree related logic in here. As commented above, LSM tree is internal to `TimelineLoader::LayoutTwoActive, LayoutTwoCompacted`, so just those 2 loaders can make use of the logic here. Btw, the correct name is LSM tree, so let's keep it in full like `lsm_tree.rs` and `LSMTree` for other applicable names ?

crates/core/src/timeline/layout.rs
codope: I agree, better suggestion overall! Few things if you can clarify:

> We also avoid calling a LSM timeline since LSM tree is just one impl of Timeline which should be hidden away from Timeline's level

Should the LSM tree logic be a separate module that the loaders use, or should the loaders themselves be the LSM tree implementations? So, in the context of `TimelineLoader::LayoutTwoActive`, would this variant contain an `LSMTree` instance, or would it implement LSM tree logic directly?

Also, should the current `Timeline::load_instants()` method remain as-is, or would you prefer a different public API? For example, another API that takes a boolean to decide whether to load archived/compacted timeline. Alternatively, we could just add another API keeping `load_instants()` as-is.

xushiyan: @codope cool! keeping lsm tree specific in `lsm_tree.rs` is good. may not need to worry about re-using it for now - it can be used by the loaders exclusively, but somewhat positioned for future re-use

Looking closer - so `load_instants()` is actually a more generic helper function, it's the using of a `selector` in one of the init functions (`new_from_storage()`) that defines the behavior of timeline's instants loading. So depends on where `load_instants()` gets called, we just need to pass in the desired `TimelineSelector`

crates/core/src/timeline/mod.rs
```diff
@@ -54,75 +61,70 @@ pub const DEFAULT_LOADING_ACTIONS: &[Action] =
...
+        self.active_loader.load_instants(selector, desc).await
     }
 
-    async fn load_instants(
+    pub async fn load_instants_with_archive(
+        &self,
         selector: &TimelineSelector,
-        storage: &Storage,
         desc: bool,
+        include_archived: bool,
```
xushiyan: I understand the current TimelineSelector has this property too, which was my initial thought. But let's try to go with simple UX for the API: callers do not need to know if they need to load archived or not, which is timeline's internal impl - they just ask for a list of instants for a given time range (everything else that a timeline selector holds):
- if no specific time given, load active instants
- if a start or end time given, and falls into archived range, then load relevant archived + active instants

Timeline loader can internally decide what to load.

The only case that we might need users to set `include_archived` is to set it `false` when they might load for a time range that could be in archived but for perf reason they choose to skip archived. This is a rare advanced usage we should be good to ignore now.

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::Result;
+use log::debug;
+use std::sync::Arc;
+
+#[derive(Debug, Clone)]
+pub enum TimelineLoader {
+    LayoutOneActive(Arc<Storage>),
+    LayoutOneArchived(Arc<Storage>),
+    LayoutTwoActive(Arc<Storage>),
+    LayoutTwoCompacted(Arc<Storage>),
```
xushiyan:
```suggestion
    LayoutTwoArchived(Arc<Storage>),
```

i wasn't thinking thoroughly. i think we should stick with `LayoutTwoArchived` like for layout 1, since it's still archived timeline, just in the form of LSM tree and compacted, which is internal impl.

crates/core/src/timeline/lsm_tree.rs
```diff
@@ -0,0 +1,70 @@
...
+pub const LSM_TIMELINE_DIR: &str = ".hoodie/timeline";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TimelineManifest {
+    pub version: i64,
+    pub entries: Vec<ManifestEntry>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ManifestEntry {
```
xushiyan, line 44: maybe some docs help explain this struct

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,111 @@
...
+                } else {
+                    Ok(instants)
+                }
+            }
+            TimelineLoader::LayoutTwoActive(storage) => {
+                let files = storage.list_files(Some(LSM_TIMELINE_DIR)).await?;
+                let mut instants = Vec::new();
+
+                for file_info in files {
+                    if file_info.name.starts_with("history/") || file_info.name.starts_with('_') {
```
xushiyan: if we can enhance the `storage.list_files()` to support a listing for only immediate files under the prefix, then we dont need explicit check and just grab the instant files. What is the case with having `_` prefix?

xushiyan: it also occurs to me that we may need to enhance the listing by ignoring `.crc` files by default.. we'll track this separately.

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,111 @@
...
+        &self,
+        _selector: &TimelineSelector,
+        _desc: bool,
+    ) -> Result<Vec<Instant>> {
+        match self {
+            TimelineLoader::LayoutOneArchived(_) => Err(CoreError::Unsupported(
+                "Archived timeline for layout v1 not implemented yet".to_string(),
+            )),
+            TimelineLoader::LayoutTwoCompacted(_) => Err(CoreError::Unsupported(
+                "Compacted timeline for layout v2 not implemented yet".to_string(),
```
xushiyan: totally ok to do archived timeline support separately. reminder to fix the name as it's still archived in V2. also this method may not be pub for hiding the impl.

crates/core/src/timeline/lsm_tree.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub file_name: String,
+    pub min_instant: String,
+    pub max_instant: String,
+    pub level: i32,
+    pub file_size: i64,
+}
+
+pub struct LSMTree {
+    storage: Arc<Storage>,
```
xushiyan, line 56: need to take any hudi configs for LSM tree specific settings?

crates/core/src/timeline/mod.rs
```diff
@@ -54,75 +61,70 @@ pub const DEFAULT_LOADING_ACTIONS: &[Action] =
...
+        // Load archived/compacted if flag is set
+        if include_archived {
+            if let Some(archived_loader) = &self.archived_loader {
+                let archived_instants = archived_loader
+                    .load_archived_instants(selector, desc)
+                    .await?;
+                instants.extend(archived_instants);
+
+                // Re-sort after merging
+                instants.sort_unstable();
```
xushiyan: can we be sure about loading results' order- archived instants already sorted right? then we can just concat with active ones?

crates/core/src/timeline/lsm_tree.rs
```diff
@@ -0,0 +1,70 @@
...
+ * under the License.
+ */
+
+use crate::error::CoreError;
+use crate::storage::Storage;
+use crate::Result;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+pub const LSM_TIMELINE_DIR: &str = ".hoodie/timeline";
```
xushiyan: we need to respect `"hoodie.timeline.history.path"`, and `"hoodie.timeline.history.path"` right? same for v6, we need to handle `"hoodie.archivelog.folder"`. but we can deal with these in PR for archived timeline. the corresponding timeline loader should be responsible to take these from configs and own it

crates/core/src/timeline/builder.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let layout_version = self
+            .hudi_configs
+            .try_get(TimelineLayoutVersion)
+            .map(|v| v.to::<isize>())
+            .unwrap_or_else(|| if table_version == 8 { 2 } else { 1 });
+
+        match layout_version {
+            1 => {
+                if !(5..=8).contains(&table_version) {
```
xushiyan: reminder- master version dropped v5 support

crates/core/src/timeline/mod.rs
codope: Agreed. We’ll keep public API simple and make archived loading an internal concern driven by the selector range; removed the public flag for now.

crates/core/src/timeline/loader.rs
codope: Ensured sorted results; removed `_` filter; ignoring `.crc`. Will do a shallow-listing enhancement on Storage in a follow-up.

crates/core/src/timeline/lsm_tree.rs
codope: We'll pass configs into the LSM reader when adding archived reads to honor timeline-specific settings.

crates/core/src/timeline/mod.rs
codope: Loaders now return sorted lists; we'll replace the final sort with a merge when archived reading lands.

crates/core/src/timeline/loader.rs
codope: Made archived loading methods non-public and aligned messages to "archived" for v2 as well.

crates/core/src/timeline/builder.rs
codope: updated builder to drop v5 support

crates/core/src/timeline/lsm_tree.rs
codope: Noted; i'll wire in the configs but archived loaders will actually resolve paths from configs in the next PR.

crates/core/src/timeline/selector.rs
```diff
@@ -523,24 +533,33 @@ mod tests {
...
+            Arc::new(HudiConfigs::new([
+                (HudiTableConfig::BasePath, "file:///tmp/base".to_string()),
+                (HudiTableConfig::TableVersion, "6".to_string()),
+            ])),
+        )
+        .unwrap();
+        let mut timeline = TimelineBuilder::new(
+            Arc::new(HudiConfigs::new([
+                (HudiTableConfig::BasePath, "file:///tmp/base".to_string()),
+                (HudiTableConfig::TableVersion, "6".to_string()),
```
xushiyan:
```suggestion
                (HudiTableConfig::BasePath, "file:///tmp/base"),
                (HudiTableConfig::TableVersion, "6"),
```

same as above. the api supports `Into<String>()`

crates/core/src/config/table.rs
```diff
@@ -166,6 +174,9 @@ impl ConfigParser for HudiTableConfig {
             Self::TimelineTimezone => Some(HudiConfigValue::String(
                 TimelineTimezoneValue::UTC.as_ref().to_string(),
             )),
+            // Fallbacks are handled in callers; no defaults here
+            Self::ArchiveLogFolder => None,
+            Self::TimelineHistoryPath => None,
```
xushiyan: these 2 do have default values right? why not provide them through this API?

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+                } else {
+                    Ok(instants)
+                }
+            }
+            TimelineLoader::LayoutTwoActive(storage) => {
+                let files = storage.list_files(Some(LSM_TIMELINE_DIR)).await?;
+                let mut instants = Vec::new();
+
+                for file_info in files {
+                    if file_info.name.starts_with("history/") || file_info.name.ends_with(".crc") {
```
xushiyan, line 188:
```suggestion
                    // TODO: make `storage.list_files` api support such filtering, like ignore crc and return files only
                    if file_info.name.starts_with("history/") || file_info.name.ends_with(".crc") {
```

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+                        continue;
+                    }
+                    match selector.try_create_instant(file_info.name.as_str()) {
```
xushiyan, line 191: i think this API needs to handle v8 instants as the completed ones have completion ts

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+        &self,
+        selector: &TimelineSelector,
+        desc: bool,
+    ) -> Result<Vec<Instant>> {
+        // Config wiring: if archived read not enabled, return behavior based on policy
+        let storage = match self {
+            TimelineLoader::LayoutOneArchived(storage)
+            | TimelineLoader::LayoutTwoArchived(storage) => storage.clone(),
+            _ => return Ok(Vec::new()), // Active loaders don't have archived parts
+        };
```
xushiyan: this storage clone does not need happen here?

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let configs: Arc<HudiConfigs> = storage.hudi_configs.clone();
+        let enabled = configs
+            .get_or_default(TimelineArchivedReadEnabled)
+            .to::<bool>();
+        if !enabled {
+            let behavior = configs
+                .get_or_default(TimelineArchivedUnavailableBehavior)
```
xushiyan: looks like we just need 1 config? `TimelineArchivedUnavailableBehavior` default to continue will just either load archived or do nothing. even for internal config, we need to keep it simple for easy maintenance

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+            };
+        }
+
+        match self {
+            TimelineLoader::LayoutOneArchived(storage) => {
+                // Resolve archive folder from configs or fallback
+                let archive_dir = configs
+                    .try_get(ArchiveLogFolder)
+                    .map(|v| v.to::<String>())
+                    .unwrap_or_else(|| ".hoodie/archived".to_string());
```
xushiyan: here this default value can be provided through `default_value()` ? the `try_get()` API handles it

crates/core/src/timeline/loader.rs
```diff
@@ -0,0 +1,172 @@
...
+
+                // List files and try creating instants through selector
+                let files = storage.list_files(Some(&archive_dir)).await?;
+                let mut instants = Vec::new();
+                for file_info in files {
+                    if file_info.name.ends_with(".crc") {
+                        continue;
+                    }
```
xushiyan: i think we can skip checking `.crc` in this pr for simplicity. there is an open PR addresses this from storage level

crates/core/src/timeline/loader.rs
xushiyan: @codope this is the only blocker for merging. if we need to pass v8 timeline loading test cases, this parsing needs to be there.

crates/core/src/config/table.rs
codope: yes, updated.

crates/core/src/timeline/loader.rs
codope: used `get_or_default`

xushiyan (approved):

Issue #367, opened 2025-06-28
Adopt stream pattern for StorageReader and related IO


xushiyan:
Use `GetResult::into_stream()` for full async IO at the storage layer

PR #411, opened 2025-08-02
feat: add `get_file_slices_splits_between` API

### LINKED ISSUE #403: Add API `get_file_slices_splits_between()`
Just like `get_file_slices_splits_as_of()` for time-travel, the incremental read API `get_file_slices_between()` should have a variant to get splits like `get_file_slices_splits_between()` that respect the read parallelism config. It should use the same splitting flow as the time-travel API.

yunchipang:
## Description

Add `get_file_slices_splits_between()` API, enables parallel processing of incremental data changes between timestamps.

closes #403

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/crates/core/src/table/mod.rs
+++ b/crates/core/src/table/mod.rs
@@ -520,6 +520,72 @@ impl Table {
             })
     }
 
+    /// Get all the changed [FileSlice]s in splits from the table between the given timestamps.
+    ///
+    /// # Arguments
+    ///     * `n` - The number of chunks to split the file slices into.
+    ///     * `start_timestamp` - If provided, only file slices that were changed after this timestamp will be returned.
+    ///     * `end_timestamp` - If provided, only file slices that were changed before or at this timestamp will be returned.
+    ///
+    /// # Notes
+    ///     * This API is useful for implementing incremental query with read parallelism.
+    ///     * Uses the same splitting flow as the time-travel API to respect read parallelism config.
+    pub async fn get_file_slices_splits_between(
+        &self,
+        n: usize,
+        start_timestamp: Option<&str>,
+        end_timestamp: Option<&str>,
+    ) -> Result<Vec<Vec<FileSlice>>> {
+        // If the end timestamp is not provided, use the latest commit timestamp.
+        let Some(end) =
+            end_timestamp.or_else(|| self.timeline.get_latest_commit_timestamp_as_option())
+        else {
+            // No latest commit timestamp means the table is empty.
+            return Ok(Vec::new());
+        };
+
+        let start = start_timestamp.unwrap_or(EARLIEST_START_TIMESTAMP);
+
+        self.get_file_slices_splits_between_internal(n, start, end)
+            .await
+    }
+
+    /// Same as [Table::get_file_slices_splits_between], but blocking.
+    pub fn get_file_slices_splits_between_blocking(
+        &self,
+        n: usize,
+        start_timestamp: Option<&str>,
+        end_timestamp: Option<&str>,
+    ) -> Result<Vec<Vec<FileSlice>>> {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()?
+            .block_on(async {
+                self.get_file_slices_splits_between(n, start_timestamp, end_timestamp)
+                    .await
+            })
+    }
+
+    async fn get_file_slices_splits_between_internal(
+        &self,
+        n: usize,
+        start_timestamp: &str,
+        end_timestamp: &str,
+    ) -> Result<Vec<Vec<FileSlice>>> {
+        let file_slices = self.get_file_slices_between_internal(start_timestamp, end_timestamp).await?;
+        if file_slices.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let n = std::cmp::max(1, n);
+        let chunk_size = file_slices.len().div_ceil(n);
+
+        Ok(file_slices
+            .chunks(chunk_size)
+            .map(|chunk| chunk.to_vec())
+            .collect())
+    }
+
     async fn get_file_slices_between_internal(
         &self,
         start_timestamp: &str,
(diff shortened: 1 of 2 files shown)


crates/core/src/table/mod.rs
```diff
@@ -520,6 +520,72 @@ impl Table {
...
+            return Ok(Vec::new());
+        }
+
+        let n = std::cmp::max(1, n);
+        let chunk_size = file_slices.len().div_ceil(n);
+
+        Ok(file_slices
+            .chunks(chunk_size)
+            .map(|chunk| chunk.to_vec())
+            .collect())
```
xushiyan: the splitting logic is repeated and gonna be the same for any new APIs in future. Would you refactor it as a util function? it can be put in `util::collection`

crates/core/src/table/mod.rs
```diff
@@ -1260,6 +1326,58 @@ mod tests {
...
     }
 
+    #[test]
+    fn empty_hudi_table_get_file_slices_splits_between() {
+        let base_url = SampleTable::V6Empty.url_to_cow();
+        let hudi_table = Table::new_blocking(base_url.path()).unwrap();
+        let file_slices_splits = hudi_table
+            .get_file_slices_splits_between_blocking(2, Some(EARLIEST_START_TIMESTAMP), None)
+            .unwrap();
+        assert!(file_slices_splits.is_empty())
```
xushiyan, line 1319: it's weird that the coverage report says missing coverage on the empty case but it has a UT covering it. maybe you can try look into that a bit further to understand why? see if it's just some codecov glitch then we can ignore

crates/core/src/table/mod.rs
```diff
@@ -520,6 +520,72 @@ impl Table {
             })
     }
 
+    /// Get all the changed [FileSlice]s in splits from the table between the given timestamps.
+    ///
+    /// # Arguments
+    ///     * `n` - The number of chunks to split the file slices into.
+    ///     * `start_timestamp` - If provided, only file slices that were changed after this timestamp will be returned.
+    ///     * `end_timestamp` - If provided, only file slices that were changed before or at this timestamp will be returned.
```
codope, line 519: Is it the same semantics (start_timestamp exclusive, end_timestamp inclusive) across other engines? Start inclusive, end exclusive feels more intuitive but if our semantics match across engines, I am fine with it.

crates/core/src/table/mod.rs
codope: +1

crates/core/src/table/mod.rs
```diff
@@ -520,6 +520,72 @@ impl Table {
...
+        let start = start_timestamp.unwrap_or(EARLIEST_START_TIMESTAMP);
+
+        self.get_file_slices_splits_between_internal(n, start, end)
+            .await
+    }
+
+    /// Same as [Table::get_file_slices_splits_between], but blocking.
+    pub fn get_file_slices_splits_between_blocking(
+        &self,
+        n: usize,
```
codope: Consider renaming parameter `n` to `num_splits` for readability in public APIs. And how does n=0 behave? Does it return a single split containing all elements?

crates/core/src/table/mod.rs
```diff
@@ -520,6 +520,72 @@ impl Table {
...
+        start_timestamp: Option<&str>,
+        end_timestamp: Option<&str>,
+    ) -> Result<Vec<Vec<FileSlice>>> {
```
codope, line 550: Returning `Vec<Vec<FileSlice>>` matches existing APIs but clones. This is fine now and consistent; if you need to optimize later, a slice-based variant could be introduced without breaking the current API. We need a util that takes `Vec<FileSlice>` and returns `Vec<Vec<FileSlice>>` by moving, not cloning.

crates/core/src/table/mod.rs
```diff
@@ -1260,6 +1326,58 @@ mod tests {
         assert!(file_slice_2.log_files.is_empty());
     }
 
+    #[test]
```
codope, line 1312: let's also add a couple of boundary conditions like n=0, end_timestamp=None.

crates/core/src/table/mod.rs
xushiyan: @codope yes i wanted to go with `[start, end)` the more common semantics, but then i realized that we have to make end inclusive for getting file slices for fresher data. say you set end = latest commit time, but you don't return the latest records until the next pull, which can make people annoyed :). We have end inclusive also makes it aligned with other query types semantics: when end is omitted, it's implicitly getting the latest, which aligns with snapshot query (T_end = the latest inclusive); when end is present, it's getting the data up to T_end, aligning with time-travel query.

crates/core/src/table/mod.rs
xushiyan: this is handled in the corresponding `xxx_internal()` shared private API

xushiyan: yup agree naming change

crates/core/src/table/mod.rs
xushiyan: yes this was more of a placeholder return type for streaming style to be adopted in future. A related issue https://github.com/apache/hudi-rs/issues/367

Ideally we should have a stream of `Vec<FileSlice>` getting consumed lazily. We need to implement it :)

crates/core/src/table/mod.rs
xushiyan: yup that can be handled in the extracted common helper function for splitting

crates/core/src/table/mod.rs
xushiyan: this PR is more of repeating the existing API pattern, we can keep changes to 
- extract split util function with UT coverage
- n renaming across all related APIs (ok to be a separate PR)

@yunchipang will you be able to make these changes?

crates/core/src/table/mod.rs
xushiyan: end time = None case has coverage see in `mod test_incremental_queries`

crates/core/src/table/mod.rs
yunchipang: @xushiyan @codope thanks for the review! and sure I will take a look and make edits asap.

crates/core/src/table/mod.rs
yunchipang: fixed!

crates/core/src/table/mod.rs
codope: @yunchipang I think we can do it in a followup so that we fix for all the APIs at once. This PR should just focus on the new API following the existing semantics.

crates/core/src/table/mod.rs
yunchipang: thanks for the suggestion. moved!

crates/core/src/table/mod.rs
codope: Would you like to pick up https://github.com/apache/hudi-rs/issues/367?

crates/core/src/table/mod.rs
codope: makes sense now.

codope (approved):

crates/core/src/table/mod.rs
yunchipang: I've fixed n naming within the `mod.rs` scope, will create an issue for renaming it across all related APIs.
and sure I can take up #367 (if it's not an urgent one ><)

xushiyan (approved):

PR #412, opened 2025-08-04
feat: expose datafusion custom table provider to python

brayanjuls:
## Description

This work seeks to expose the existing datafusion datasource to the python api, with the goal of being able to use dafusion in python hudi-rs. To implement this we leverage datafusion FFI custom table provider as mentioned in this [doc](https://github.com/apache/datafusion-python/blob/main/docs/source/user-guide/io/table_provider.rst). 

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/python/hudi/_internal.pyi
+++ b/python/hudi/_internal.pyi
@@ -349,3 +349,18 @@ def build_hudi_table(
         HudiTable: An instance of hudi table.
     """
     ...
+@dataclass(init=False)
+class HudiDataSource:
+    def __init__(
+        self,
+        base_uri: str,
+        options: Optional[List[Tuple[str, str]]] = None,
+    ):
+        """
+        Initializes the HudiDataSource.
+
+        Parameters:
+            base_uri (str): The base URI of the Hudi table.
+            options (Optional[List[Tuple[str, str]]]): Additional configuration options (optional).
+        """
+        ...
--- a/python/src/internal.rs
+++ b/python/src/internal.rs
@@ -18,9 +18,15 @@
  */
 
 use arrow::pyarrow::ToPyArrow;
+use datafusion::error::DataFusionError;
+use datafusion_ffi::table_provider::FFI_TableProvider;
+use hudi::HudiDataSource as InternalHudiDataSource;
+use pyo3::{types::PyCapsule, Bound};
 use std::collections::HashMap;
 use std::convert::From;
+use std::ffi::CString;
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::sync::OnceLock;
 use tokio::runtime::Runtime;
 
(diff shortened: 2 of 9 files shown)


brayanjuls:
hi @xushiyan, I don't have an issue associated for this PR if you need it I can go ahead and create one.

xushiyan:
> hi @xushiyan, I don't have an issue associated for this PR if you need it I can go ahead and create one.

@brayanjuls thanks for the PR! It's alright we don't need to have an issue for this.

brayanjuls:
I see the overall coverage decreases because of my changes on internal.rs, do I need to add unit testing there or it is ok to only have unit testing in python?

xushiyan:
> I see the overall coverage decreases because of my changes on internal.rs, do I need to add unit testing there or it is ok to only have unit testing in python?

@brayanjuls we evaluate the coverage case by case. if it's non-critical, and the CI shows green, then we're good.

python/src/internal.rs
```diff
@@ -239,6 +251,48 @@ impl From<&Instant> for HudiInstant {
     }
 }
 
+#[cfg(not(tarpaulin_include))]
+#[pyclass(name = "HudiDataSource")]
+#[derive(Clone)]
+pub struct HudiDataSource {
+    table: InternalHudiDataSource,
```
xushiyan: in rust, we have hudi-datafusion crate, so HudiDataSource is meant for datafusion. 
In python, we need to make the name explicit for datafusion so people don't confuse it with datasource for other integrations (in future). 
Also can we make all related functionalities in this PR available only when the python `datafusion` feature is enabled? like people use `pip install hudi[datafusion]` to access the related features.

python/hudi/_internal.pyi
```diff
@@ -349,3 +349,18 @@ def build_hudi_table(
         HudiTable: An instance of hudi table.
     """
     ...
+@dataclass(init=False)
+class HudiDataSource:
```
xushiyan: would love to see this refactored into another `.pyi` for better code organization. it'll be optional here and you can do it separately if you like.

xushiyan (commented):
Thanks for the nice helpful contribution!

python/src/internal.rs
brayanjuls: I renamed HudiDataSource to HudiDataFusionDataSource, also make datafusion dependency optional in the pyproject.toml so people can use `pip install hudi[datafusion]`.

Additionally I extracted HudiDataFusionDataSource(datafusion_internal.rs) to a different module so we can enable or disable it with rust feature flags.

python/hudi/_internal.pyi
brayanjuls: I tried to do this but got some errors, so I think I will try to do it in a different PR.

xushiyan:
> @xushiyan do you need any additional changes from me in this PR?

being away for some work stuff. LGTM.

PR #432, opened 2025-08-30
refactor: replace custom `.to::()` with `into()` for `HudiConfig` conversions

### LINKED ISSUE #368: Use `into()` for HudiConfig conversion to primitive types
Replace the custom API of conversions like `.to::<String>()`.

yunchipang:
## Description

- delete custom `.to::<T>()` method from `HudiConfigValue`
- replaced all `.to::<Type>()` calls with standard `.into()` patterns across the codebase

closes #368

- [ ] N/A
- [ ] Manual tests

### DIFF
--- a/crates/core/src/config/read.rs
+++ b/crates/core/src/config/read.rs
@@ -142,20 +142,26 @@ mod tests {
             ),
         ]);
         assert_eq!(
-            InputPartitions.parse_value(&options).unwrap().to::<usize>(),
+            { let value: usize = InputPartitions.parse_value(&options).unwrap().into(); value },
             100
         );
         assert_eq!(
-            ListingParallelism
-                .parse_value(&options)
-                .unwrap()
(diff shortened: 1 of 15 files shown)


crates/core/src/config/read.rs
```diff
@@ -142,20 +142,26 @@ mod tests {
             ),
         ]);
         assert_eq!(
-            InputPartitions.parse_value(&options).unwrap().to::<usize>(),
+            { let value: usize = InputPartitions.parse_value(&options).unwrap().into(); value },
```
xushiyan: try to avoid the pattern. use sth like this before the assert to keep code clean. `actual` can be reused to save naming hassle. applies to all other similar changes

```rust
        let actual: usize = InputPartitions.parse_value(&options).unwrap().into();
```

yunchipang: fixed!

PR #433, opened 2025-09-02
feat: export `HudiTableConfig` and `HudiReadConfig` as python enums

### LINKED ISSUE #404: Export `HudiTableConfig` and `HudiReadConfig` as enum in Python APIs
`HudiTableConfig` and `HudiReadConfig` should be made available in Python APIs

yunchipang:
## Description

Expose `HudiTableConfig` and `HudiReadConfig` as native Python `enum.Enum` classes, reflected from the Rust core at module-import time.

- Adds a small `_config_keys` pyfunction that returns `(variant_name, key_string)` pairs for each config enum, using `strum::IntoEnumIterator` and `IntoStaticStr`.
- `python/hudi/_config.py` builds the Python enums via `Enum(...)` functional API with `str` mixin, so members compare equal to their underlying key strings and are hashable, iterable, and support name/value lookup.
- `HudiTableBuilder.with_hudi_option` and `with_option` accept either a string or an enum member; existing string-based usage is unchanged.
- Variant names and key strings have a single source of truth in Rust — adding a new Rust variant requires no Python code changes.

closes #404

- [ ] N/A
- [ ] Manual tests


python/hudi/_internal.pyi
```diff
@@ -15,12 +15,45 @@
...
+    POPULATES_META_FIELDS = "hoodie.populate.meta.fields"
+    RECORD_KEY_FIELDS = "hoodie.table.recordkey.fields"
+    RECORD_MERGE_STRATEGY = "hoodie.table.record.merge.strategy"
+    TABLE_NAME = "hoodie.table.name"
+    TABLE_TYPE = "hoodie.table.type"
+    TABLE_VERSION = "hoodie.table.version"
+    TIMELINE_LAYOUT_VERSION = "hoodie.timeline.layout.version"
+    TIMELINE_TIMEZONE = "hoodie.table.timeline.timezone"
+
+class HudiReadConfig(Enum):
```
xushiyan: the goal is to auto expose the config enums so we don't have to copy the config strings for another language binding. try to research if this is doable with pyo3 to bind python enum to a rust enum without duplicating every config entry.

yunchipang: @xushiyan i added a `macros` crate to auto-gen enums for python using pyo3. PTAL if this aligns better with the goal. thanks!

xushiyan (approved):

PR #457, opened 2025-09-22
fix: replace Windows backslashes with forward slashes in paths

### LINKED ISSUE #445: Windows path separator bug: library uses backslash instead of forward slash, causing issues for cloud and POSIX APIs
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of the bug

On Windows, the library uses `PathBuf` for joining paths, which results in paths with backslashes (`\`). This causes problems when interacting with cloud storage APIs or POSIX-style APIs that expect forward slashes (`/`) as the path separator. As a result, file operations may fail, or the generated paths may not be accepted by external systems. There is no documented workaround or environment variable to force forward slash normalization for cross-platform compatibility.

### Steps To Reproduce

1. Use python hudi to access some hudi table in S3 on Windows.

2. Run example code:
```python
from hudi import HudiTableBuilder
import pyarrow as pa

hudi_table = HudiTableBuilder.from_base_uri("s3://path/to_your_hudi/").build()
batches = hudi_table.read_snapshot(filters=[("city", "=", "san_francisco")])

# convert to PyArrow table
arrow_table = pa.Table.from_batches(batches)
result = arrow_table.select(["rider", "city", "ts", "fare"])
print(result)
```
3. Observe s3 error: path to .hoodie snapshot will be concatenated with backslash

### Expected behavior

Paths should always use forward slashes (`/`) when interacting with external APIs, regardless of the underlying OS. There should be a documented way (e.g., environment variable or API option) to normalize paths for cross-platform compatibility.

### Screenshots / Logs

Storage error: Object at location
my_table/.hoodie\20250909105111479.deltacommit not found: Client error with status 404 Not Found: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not
exist.</Message><Key>my_table/.hoodie\20250909105111479.deltacommit</Key><RequestId>XXXX</RequestId><HostId>YYYY</HostId></Error>

### Software information

- Operating system: Windows
- Project version: 0.4.0

### Additional context

This bug can cause major interoperability problems for users working on Windows, especially with cloud storage (S3, GCS, etc.). Please consider adding a normalization step or a config option/environment variable to always use POSIX-style paths when interacting with external systems.

yunchipang:
## Description

Write util functions to replace backslashes with forward slashes when constructing file paths.

fixes: #445 

- [ ] N/A
- [x] Manual tests (OS: Windows 11, Python: 3.13.7, in real AWS S3 env)

### DIFF
--- a/crates/core/src/storage/util.rs
+++ b/crates/core/src/storage/util.rs
@@ -18,7 +18,8 @@
  */
 
 //! Utility functions for storage.
-use object_store::path::Path as ObjectPath;
+
+use std::path::PathBuf;
 use url::Url;
 
 use crate::storage::error::StorageError::{InvalidPath, UrlParseError};
@@ -28,25 +29,7 @@ use crate::storage::Result;
 pub fn parse_uri(uri: &str) -> Result<Url> {
     let mut url = match Url::parse(uri) {
         Ok(url) => url,
-        Err(e) => {
-            let path = std::path::Path::new(uri);
-
-            let absolute_path = if path.is_absolute() {
-                path.to_path_buf()
-            } else {
-                std::env::current_dir()
-                    .map_err(|_| UrlParseError(e))?
-                    .join(path)
-            };
-
-            let final_path = if absolute_path.exists() {
-                absolute_path
-            } else {
-                absolute_path.canonicalize().unwrap_or(absolute_path)
-            };
-
-            Url::from_file_path(&final_path).map_err(|_| UrlParseError(e))?
-        }
+        Err(e) => Url::from_directory_path(uri).map_err(|_| UrlParseError(e))?,
     };
 
     if url.path().ends_with('/') {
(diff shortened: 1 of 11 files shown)


crates/core/src/storage/util.rs
```diff
@@ -43,6 +43,22 @@ pub fn get_scheme_authority(url: &Url) -> String {
...
+
+/// Joins path segments with forward slashes.
+pub fn join_storage_path(segments: &[&str]) -> String {
+    let joined = segments
+        .iter()
+        .map(|s| s.trim_matches('/').trim_matches('\\'))
+        .filter(|s| !s.is_empty())
+        .collect::<Vec<_>>()
+        .join("/");
+    normalize_path_for_storage(&joined)
```
xushiyan: we don't want to maintain such conversion. see if we can leverage object store API which should have already handled the paths. see https://docs.rs/object_store/latest/object_store/path/struct.Path.html#method.from_filesystem_path

basically we need to handle 1) local fs paths concatenation which rust's std::path should work 2) for cloud store paths, we always handle it with object store APIs

xushiyan: @yunchipang i've cherrypicked the commit with CI test setup for windows here so it can detect more issues wrt the change. let's fix the PR to have CI all green. then i'll revert the CI commit before merging it.

yunchipang: @xushiyan okie. just pushed an update could you trigger CI again? thanks

PR #487, opened 2025-11-27
ci: optimize integration test CI time

### LINKED ISSUE #270: Improve integration test CI time
Consider optimization like skipping running integ test when the PR is indicated as `docs` or `chore`. And consider split the tasks.

AndreaBozzo:
Reduces integration test CI time by using a pre-built Docker image from Docker Hub.

## Approach

Per maintainer suggestion, the runner image is now pre-built and published to Docker Hub instead of building it in CI:

- **Image**: [andreab92/hudi-demo-runner:latest](https://hub.docker.com/r/andreab92/hudi-demo-runner)
- **Tags**: rust-1.87-py3.13, latest

## Changes

- Remove prepare-integration-build job
- Pull pre-built image directly from Docker Hub
- Use GitHub Actions cache for Rust/Python build artifacts
- Revert ci_run.sh to original simple version
- Restore networks section in compose.yaml

## After merge

Maintainer can republish under official Hudi Docker Hub.

Fixes #270

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -31,7 +31,7 @@ jobs:
   check-pr:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Setup Node
         uses: actions/setup-node@v6


AndreaBozzo:
Fixed the CI failures - three issues were blocking the workflow:

1. **UV PATH not found** - `docker compose exec` wasn't inheriting the container's PATH, so the `uv` binary wasn't accessible. Added explicit PATH export.

2. **Cache rotation failing on first run** - The buildx cache didn't exist yet, causing `mv` to fail. Added a check before moving.

3. **Labeler permissions** - Fork PRs can't use `pull_request` trigger with write permissions. Created a separate `pr-labeler.yml` workflow using `pull_request_target` (which is safe because it only checks out the base branch, not the PR code).

Testing the labeler workflow on this PR to make sure it works correctly, this might take a few tries xd

demo/infra/runner/Dockerfile
```diff
@@ -36,6 +37,10 @@ ENV PATH="/opt/.venv/bin:$PATH"
 
 RUN uv pip install --only-binary=all 'pyarrow==20.0.0'
 
+# Make Python installation and venv accessible to any user (needed for HOST_UID:HOST_GID mapping)
+# Python needs read+execute, venv needs read+write+execute for pip install
+RUN chmod -R a+rX /opt/python && chmod -R a+rwX /opt/.venv
```
xushiyan, line 42: can you pre-build this docker image and publish it in docker hub so it can be pulled directly during CI? once everything passed, I can re-publish the same image under Hudi's official docker hub

.github/workflows/ci.yml
```diff
@@ -49,6 +49,129 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  prepare-integration-build:
```
xushiyan: this job currently adds 7min overhead before running integ test. pls see how to optimize. for eg, if you have pre-built image, then no need to pull base and prepare. the pre-built image should be directly used for the docker compose stack.

.github/workflows/ci.yml
```diff
@@ -200,7 +333,7 @@ jobs:
     if: github.event.pull_request.user.login != 'dependabot[bot]'
     steps:
       - uses: actions/checkout@v6
-      - uses: actions/download-artifact@v5
+      - uses: actions/download-artifact@v4
```
xushiyan: why downgrade?

.github/workflows/pr.yml
```diff
@@ -46,5 +46,4 @@ jobs:
 
       - name: Labeling
         uses: actions/labeler@v6
-        # disable until figuring out a way to do this without pull_request_target
-        if: false
+        if: false  # Temporarily disabled - permissions issue
```
xushiyan: revert this change. we don't touch labeler for now

.gitignore
```diff
@@ -19,6 +19,10 @@
 /target
 **/target
 
+# CI/Docker cache directories
+.cargo/
+.uv-cache/
+
```
xushiyan: why would this matter for CI? also we are keeping `.cargo/config.toml` for the repo

demo/compose.yaml
```diff
@@ -70,6 +71,3 @@ services:
       AWS_ALLOW_HTTP: true
       AWS_REGION: us-east-1 # minio default
 
-networks:
-  demo_network:
-    driver: bridge
```
xushiyan, line 75: what is the impact of removing this network setup?

demo/infra/runner/Dockerfile
AndreaBozzo: I'm going to try do that before Christmas, thats actually neat solution

demo/compose.yaml
AndreaBozzo: Defined but seemed unassigned to any service, docker compose creates a network by default, removing shouldnt change this behaviour i think? 
i can revert that aswell

demo/infra/runner/Dockerfile
xushiyan: sounds good

.github/workflows/pr.yml
```diff
@@ -31,7 +31,7 @@ jobs:
   check-pr:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
```
xushiyan: revert

demo/compose.yaml
xushiyan: this isolates demo stack's network from others

demo/ci_run.sh
```diff
@@ -25,7 +25,35 @@ export COMPOSE_DOCKER_CLI_BUILD=1
...
+# Check for cached build flag (pre-compiled artifacts available in mounted volume)
+USE_CACHED_BUILD=false
+if [ "$2" = "--use-cached-build" ]; then
+  USE_CACHED_BUILD=true
+fi
+
+# Start services with or without rebuild
+if [ "$USE_CACHED_BUILD" = "true" ]; then
+  # In CI, the runner image is pre-built and tagged as demo-runner:cached
+  # Note: we still need to build mc (minio client) as it's not cached, but it's small
```
xushiyan: you probably want to look into those github action steps like download artifact which may already handle caching so that we don't need to maintain caching logic outside GH workflows. the goal is to minimize the script for easy maintenance

AndreaBozzo:
> @AndreaBozzo i'll investigate later easier for me to trigger CI and see how it works. meanwhile can you rebase and resolve conflict ?

updated branch, don't worry i have no rush and thanks again!

PR #639, opened 2026-07-31
feat(core): add the merge-on-read reader context, resolver, and entry point

linliu-code:
## What

Lays the groundwork for the merge-on-read file group reader, in three parts:

1. **`ReaderContext` + resolver** — everything the reader needs that is derived from table state, and the code that derives it from a table's configs.
2. **An equivalence test** — pins that the resolver derives the same instant range the existing file group reader does, so the two cannot drift.
3. **`FileGroupReader` entry point** — the shape callers will use. `read` returns `Unsupported` for now.

Nothing consumes any of it: `reader_v2` is `pub(crate)` and unreferenced. **Public API delta: none.**

## Why this first

This is the first change in a sequence that ports the merge-on-read file group reader. Resolving the context up front gives every later change a defined target to build against, and concentrates the read-semantics decisions in one small reviewable place rather than scattering them across the port.

## Decisions that need review

**`append_only` is rejected rather than mapped.** The merge-on-read reader always merges by record key, so no merge mode reproduces "keep every version". The table config derives `append_only` whenever meta fields are disabled or no ordering field is set, which makes it reachable for ordinary tables — so guessing a mode here would silently change which rows a query returns. This is also why `MergeMode::CommitTimeOrdering` is currently unconstructible; the variant documents what makes it reachable. See `resolver::resolve_merge_mode`.

**`latest_commit_time` is required, not defaulted.** It bounds which log blocks are visible. A `FileGroupReader` built straight from a base URI never loads the timeline, so the caller must supply it — defaulting would silently widen the read.

**`should_merge_use_record_position` stays off.** Log blocks already carry record positions but nothing reads them, so merging by key is what a read does today.

## Notable

`FileGroupReader::create_instant_range_for_log_file_scan` is widened from private to `pub(crate)` so the equivalence test can reach it. Visibility only — no behavior change, and its two existing call sites are untouched.

The new type shares a name with the existing `FileGroupReader`, which it replaces once the engine behind it is written. It carries the name it will keep, so the eventual swap is a module deletion rather than a rename touching every call site; until then the two are told apart by module path.

## Deferred

The schema handler, the record context, and the bootstrap flags — until something reads them. `merge_strategy_id` and `iterator_mode` are omitted too: the merger dispatches on the merge mode, and only one iterator mode is implemented.

## Testing

16 unit tests, written test-first. The two non-trivial assertions were checked by mutation — the equivalence test was confirmed to catch both a flipped range bound and a dropped start timestamp, and the not-implemented test was confirmed to fail when asserting the wrong message.

Full workspace suite green (903 tests); `cargo fmt --check` clean; no clippy findings in the new module.

### DIFF
--- a/crates/core/src/file_group/mod.rs
+++ b/crates/core/src/file_group/mod.rs
@@ -25,6 +25,7 @@ pub mod builder;
 pub mod file_slice;
 pub mod log_file;
 pub mod reader;
+pub(crate) mod reader_v2;
 pub mod record_batches;
 
 use crate::Result;
--- a/crates/core/src/file_group/reader.rs
+++ b/crates/core/src/file_group/reader.rs
@@ -177,7 +177,9 @@ impl FileGroupReader {
         apply_commit_time_filter(&self.hudi_configs, records)
     }
 
-    fn create_instant_range_for_log_file_scan(&self) -> Result<InstantRange> {
+    /// Visible to the crate so the merge-on-read context resolver can assert it
+    /// derives the same range; see `reader_v2::resolver`.
+    pub(crate) fn create_instant_range_for_log_file_scan(&self) -> Result<InstantRange> {
         let timezone = self
             .hudi_configs
             .get_or_default(HudiTableConfig::TimelineTimezone)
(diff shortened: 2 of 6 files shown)


linliu-code:

PR #640, opened 2026-08-02
feat(core): port the reader's stats, profiling, and input-split leaves

linliu-code:
**Stacked on #639** — that PR's commits appear here until it merges. **Review only the last commit** (`port the reader's stats, profiling, and input-split leaves`).

## What

Four leaf modules of the merge-on-read reader, ported as-is: `read_stats`, `profiling`, `iterator_mode`, `input_split`. They have no dependencies inside the reader subsystem, so they need no adaptation.

Nothing consumes them — `reader_v2` stays `pub(crate)` and unreferenced, and the existing read path is untouched. **Public API delta: none.**

## Note on dead_code

Ported files carry a file-scope `allow(dead_code)`; every item in them is live upstream, so a per-item allow would be noise. Hand-written items (in #639) keep per-item allows so a mistakenly-dead one still warns.

Build warning-free; suite green.

### DIFF
--- a/crates/core/src/file_group/mod.rs
+++ b/crates/core/src/file_group/mod.rs
@@ -25,6 +25,7 @@ pub mod builder;
 pub mod file_slice;
 pub mod log_file;
 pub mod reader;
+pub(crate) mod reader_v2;
 pub mod record_batches;
 
 use crate::Result;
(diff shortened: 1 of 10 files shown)


linliu-code:

PR #641, opened 2026-08-02
feat(core): port the reader's delete context and output converter

linliu-code:
**Stacked on #639 and #640** — their commits appear here until they merge. **Review only the last commit.**

Two more leaf modules of the merge-on-read reader: `delete_context` (how a read recognizes a delete — the marker field/value a payload uses) and `output_converter` (projects a merged batch to the requested columns).

Nothing consumes them; `reader_v2` stays `pub(crate)` and unreferenced. **Public API delta: none.** Build warning-free, suite green.


linliu-code:

PR #642, opened 2026-08-02
feat(core): port Avro schema comparison and Parquet list normalization

linliu-code:
**Stacked on #639, #640, #641** — their commits appear here until they merge. **Review only the last commit.**

Two schema helpers the merge-on-read reader depends on:

- `schema::avro_schema_utils` — decides whether two Avro schemas are projection-equivalent, so the reader can tell a genuine schema change from a reordering or a narrowing that needs no work.
- `schema::parquet_list_norm` — normalizes the legacy two-level repeated-group and modern three-level Parquet list encodings, so batches from files written by different engines line up.

Also brings the two parquet fixtures `parquet_list_norm`'s tests read. They are force-added past the `**/data` ignore, the same way the existing fixtures under `crates/test/data` are tracked.

Unlike the `reader_v2` modules these are `pub` under `schema`, matching their upstream placement — nothing calls them yet. Full workspace suite green.


linliu-code:

PR #643, opened 2026-08-02
feat(core): port schema evolution for merged batches

linliu-code:
**Stacked on #639–#642** — their commits appear here until they merge. **Review only the last commit.**

Two modules that let a read span a schema change — the capability `hudi-rs` has no equivalent of today:

- `schema::extended_promotion` — which type changes a read can absorb (int→long, decimal scale widening, …), rejecting the rest rather than producing a silently wrong value.
- `schema::batch_evolution` — applies a target schema to a batch read under an older one: back-fill added columns as null, drop removed ones, cast promoted ones through the rules above.

Together they are what makes a MOR slice readable when its base file and log blocks were written under different schemas.

Nothing calls them yet. Build warning-free; 795 lib tests green.


linliu-code:

PR #644, opened 2026-08-02
feat(core): port row serialization and record positions

linliu-code:
**Stacked on #639–#643** — their commits appear here until they merge. **Review only the last commit.**

- `reader_v2::row_serde` — the Arrow IPC pair the merge buffer uses to move a record to disk and back when it spills, plus the body-only variant that drops per-record schema framing.
- `reader_v2::buffer::record_positions` — decodes the positions a log block carries for its records.

## New dependencies — the one thing to scrutinize here

`base64` and `roaring`. Hudi stores record positions in a log block header as a **base64-encoded roaring bitmap**, so these are the storage format's own encoding rather than a design choice. Neither pulls a transitive tree of its own.

Nothing reads positions yet — the position-based merge that uses them lands later. Build warning-free; 804 lib tests green.


linliu-code:

PR #645, opened 2026-08-02
feat(core): port the buffered record

linliu-code:
**Stacked on #639–#644** — their commits appear here until they merge. **Review only the last commit.**

The unit the merge works over: one record's key, its ordering value, and either its payload or the fact that it is a delete. Carries the ordering comparison the merge uses to pick a winner between two versions of a key, including the decimal rescaling that needs.

This is the type most of the remaining reader depends on, so it lands on its own.

**Note:** ported against the module shape from onehouseinc/hudi-rs-internal#112, which moved spill serialization out of the record context so the two stop importing each other. That is why this depends on `row_serde` (#644) rather than on `record_context`.

Build warning-free; 818 lib tests green.


linliu-code:

PR #646, opened 2026-08-02
feat(core): port the record merger, update processor, and record converter

linliu-code:
**Stacked on #639–#645** — their commits appear here until they merge. **Review only the last commit.**

- `record_merger` — picks a winner between two versions of a record key, in either commit-time ordering (later commit wins) or event-time ordering (higher ordering value wins). Deletes resolve through the same comparison, so a delete only wins if it is genuinely newer.
- `update_processor` — counts what a merge did (inserts / updates / deletes), which is what the read stats report.
- `buffered_record_converter` — the seam for turning an engine record into a buffered one.

This is the piece `hudi-rs` currently approximates with a whole-batch `lexsort`; it merges per record instead.

Build warning-free; 836 lib tests green.


linliu-code:

PR #647, opened 2026-08-02
feat(core): port the record context and row extraction

linliu-code:
**Stacked on #639–#646** — their commits appear here until they merge. **Review only the last commit.**

- `record_context` — reads what a merge needs out of a batch: record key, the ordering value for a row, whether a row is a delete. Resolves field names from table config once, then works per row.
- `buffer::row_extraction` — pulls a single row out of a batch as a standalone batch, which is how a merged record leaves the buffer.

## Two adaptations, both because this repo is ahead of the source

1. **Ordering fields.** Resolves `hoodie.table.ordering.fields` first with the deprecated `hoodie.table.precombine.field` still honored — upstream checks the deprecated key first, predating #598's rename.
2. **Avro schema in a test.** Upstream builds an Arrow schema from Avro JSON via the forked `arrow-avro`; here it goes through this crate's own `avro_to_arrow::to_arrow_schema`, same shape, no forked dependency.

That second one matters beyond this PR: it is evidence the reader can be ported without the forked arrow stack, which was the main open risk in this migration.

Build warning-free; 878 lib tests green.


linliu-code:

PR #648, opened 2026-08-02
feat(core): port the file group reader schema handler

linliu-code:
**Stacked on #639–#647** — their commits appear here until they merge. **Review only the last commit.**

Works out the three schemas a read needs: what the caller asked for, what must be read from storage to satisfy it, and what the merged output looks like. The required schema is wider than the requested one whenever the merge needs a column the caller did not ask for — an ordering field, say.

Also adds `schema::resolver::avro_json_to_arrow_schema`, which the handler needs to turn a log block's Avro schema into an Arrow one. It goes through this crate's own `avro_to_arrow` rather than the forked `arrow-avro` the source uses — more evidence the port does not need the forked arrow stack.

## A gap worth knowing about

The source switched *away* from the vendored converter for a reason, and it applies here. `avro_to_arrow::schema` (line 73-76) gives a list's element field `nullable = false` **unconditionally**, even when the Avro items schema is a `["null", T]` union. A merge-on-read slice whose arrays contain NULL elements will therefore resolve to a schema that disagrees with its data.

I did **not** fix it here: that conversion is shared with the existing read path, so changing it belongs in its own PR with a regression test rather than riding along with a port. It is recorded on the new function's doc comment so it cannot be lost.

Build warning-free; 900 lib tests green.


linliu-code:

PR #650, opened 2026-08-02
feat(core): port the record buffer and merge iterator

linliu-code:
**Stacked on #639–#649** — their commits appear here until they merge. **Review only the last commit.**

- `buffer` — defines what a record buffer is: takes log blocks, merges each record against what it already holds, hands back the result.
- `buffer::record_buffer` — the shared machinery every implementation builds on: merge map, merger, delete context, update counting, wired together.
- `merge_iterator` — drives the buffer, emitting the merged result **one batch at a time** rather than materializing the whole slice. This is what makes streaming MOR reads possible.

## One visibility change outside reader_v2

`log_file::log_block` goes from private to `pub`. The buffer reads block metadata directly, and the module was already public in all but declaration — its types are reachable through `log_file::reader`. No behavior change.

The re-exports in `buffer/mod.rs` for the concrete buffer implementations (`key_based`, `position_based`, `loader`) carry an `unused_imports` allow until those land.

Build warning-free; 941 lib tests green.


linliu-code:

PR #651, opened 2026-08-02
feat(core): replace the reader context with the ported one

linliu-code:
**Stacked on #639–#650** — their commits appear here until they merge. **Review only the last commit.**

## What

#639 added a `ReaderContext` written against this repo *before* the reader it serves had been ported — an approximation of the target. Now that the reader is arriving, the ported context is the real target, so this replaces the placeholder and rewires the resolver onto it.

The resolver keeps its typed `MergeMode` and converts to the string form the context carries, so the mapping stays checked where it is decided rather than becoming a bare string at the call site.

Fields with no caller in this repo resolve to inert values, each with a comment saying why:

| Field | Value | Why |
|---|---|---|
| `has_bootstrap_base_file`, `needs_bootstrap_merge` | `false` | no bootstrap support here |
| `row_filter_builder`, `mor_pk_safe` | `None` / `false` | nothing pushes predicates into the merge path |
| `completion_gate_inputs` | `None` | needs a timeline the caller has not loaded; leaves the gate a no-op |
| `merge_strategy_id` | `""` | dispatch is on `merge_mode`; carried, not consulted |
| `iterator_mode` | `ENGINE_RECORD` | the only one implemented |

Also adds `storage::RowFilterBuilder`, the type that pushdown field is typed against. Nothing installs one yet.

## Note

The instant-range equivalence test from #639 caught this change — `instant_range` became `Option` — and had to be updated. Exactly what it was written for.

Build warning-free; 949 lib tests green.


linliu-code:

PR #652, opened 2026-08-02
feat(core): port the log record readers and the key-based buffer

linliu-code:
**Stacked on #639–#651** — their commits appear here until they merge. **Review only the last commit.**

- `log_record_reader` — walks a slice's log files and decides which blocks a read admits: rollbacks remove their target, compaction blocks subsume what they compacted, the completion gate excludes instants that never finished.
- `merged_log_record_reader` — drives that scan into a buffer.
- `buffer::key_based` — the buffer that merges by record key.

## Three adaptations, one cause

Upstream reads log blocks **lazily** (windowed streaming + per-block inflate); this crate reads them **whole**. So:

1. The Pass-1 header sweep uses the existing whole-file log reader instead of a streaming one. Same block set, more memory — and no more than the existing read path already uses. Porting the streaming reader would mean **rewriting `StorageReader`, a type the existing reader shares**, which is exactly what this migration is trying not to do.
2. The scan asks for an unbounded instant range, leaving admission to the gates rather than filtering twice under different rules.
3. The two `inflate` calls are dropped — blocks arrive with content, so there is nothing to fetch.

Worth noting: upstream's `new_streaming` takes an `Arc<ReaderContext>`, which would have made the shared `log_file` layer depend on `reader_v2`. Avoiding that inversion is why this takes the eager path rather than porting the streaming one.

Also makes `log_file::log_format` public, alongside `log_block` from #650.

Build warning-free; **1076 lib tests green** (up from 949).


linliu-code:

PR #653, opened 2026-08-02
feat(core): port position-based merging and the buffer loader

linliu-code:
**Stacked on #639–#652** — their commits appear here until they merge. **Review only the last commit.**

- `buffer::position_based` — merges log records onto base rows by the **position** a record occupies in the base file rather than by key, avoiding a key index when the writer recorded positions. This capability has no equivalent in the current reader.
- `buffer::loader` — chooses between that and the key-based buffer, and runs the log scan that fills it.

Also moves `ReaderParameters` into its own module where the rest of the reader expects it, picking up the `use_record_position` flag the loader reads.

## Two additive LogBlock accessors

`record_positions_header()` and `base_file_instant_time_of_positions()`. Both headers were **already parsed** by this crate (`BlockMetadataKey::RecordPositions`, `BaseFileInstantTimeOfRecordPositions`) — they simply had no accessor, so the values were read and discarded. Purely additive; the existing reader is unaffected.

The `inflate` calls are dropped for the same reason as #652: blocks arrive with their content read.

Build warning-free; **1088 lib tests green**.


linliu-code:

PR #654, opened 2026-08-02
feat(core): port the merge-on-read file group reader

linliu-code:
**Stacked on #639–#653** — their commits appear here until they merge. **Review only the last commit.**

The reader itself, and the **last module of the port**. It resolves the schemas a read needs, opens the base file, runs the log scan into a buffer, and merges the two into batches. The subsystem is now complete — and still `pub(crate)` and unreferenced.

Base file reads go through this crate's `BaseFileReader` rather than the storage helpers upstream calls, so **Lance base files work the same way parquet ones do**.

## Reads are eager — the decision behind this PR

Upstream opens the base file as a stream and consumes it **synchronously**, one row group at a time (`block_on` per group). That deadlocks inside the async contexts this crate reads from — `Table::read_stream`, the DataFusion exec, and the Python bindings all consume from async.

So the streaming path resolves to the eager one for now: same rows, same order, higher peak memory, no worse than the existing read path. The `streaming` argument is kept so an async-native merge iterator can switch it back on without changing callers, and the streaming branch returns an error rather than panicking if reached early.

Two capabilities upstream threads through the base read are dropped and **named rather than faked**: the predicate push-down filter (no caller here) and the row-number column (position-based merging needs it; not wired up).

## Two fixes that fell out of the ported tests

- **`Instant::parse_naive_datetime` could panic.** It sliced `timestamp[..14]` without a length check, so any shorter value crashed. Callers reach it with values read from *data* — a `_hoodie_commit_time` cell — not only with instants parsed from well-formed file names. Now returns an error, matching the crate's no-panic convention.
- **`InstantRange` gains `is_in_range_lexicographic`**, which is how the reader bounds a commit time it cannot parse as a datetime. Additive; existing methods untouched.

Build warning-free; **1102 lib tests green** (709 before the port began).


linliu-code:

PR #656, opened 2026-08-03
test(core): pin that append-only reads do not apply deletes

linliu-code:
Independent of the merge-on-read reader port — this is about the **existing** read path.

## What I found

`RecordMerger`'s append-only arm is:

```rust
RecordMergeStrategyValue::AppendOnly => record_batches.concat_data_batches(self.schema.clone())
```

`concat_data_batches` concatenates `data_batches` and returns. `delete_batches` is never read. **A deleted record is still in the result.**

## Why that is reachable

Append-only is defensible for a table that genuinely only appends. But it is **not opt-in** — `config/table.rs` *derives* it:

- `hoodie.populate.meta.fields = false` → `append_only`, or
- no `hoodie.table.ordering.fields` / `precombine.field` → `append_only`

A merge-on-read table meeting either condition has log files that can carry delete blocks. Those deletes are dropped, and updates are returned as duplicates rather than merged (the existing `test_merge_records_append_only` already documents the duplication — 4 rows out of 2+2 with a repeated key).

## Why it went unnoticed

I checked every merge-on-read fixture in the repo: **all 14 set an ordering field and leave meta fields on**, so none derives append-only. The path has no fixture coverage at all.

## What this PR does

Adds a test pinning the current behavior, so a change to it is deliberate rather than accidental. It does **not** change behavior — I did not want to alter existing read results in a test-only PR, and the right fix is a design question:

- treat a merge-on-read table with delete blocks as never append-only, or
- apply deletes on the append-only path too, or
- reject the combination loudly

Worth a maintainer's call. This came up because the ported merge-on-read reader has no append-only equivalent — it always merges by key — so the same table would return different rows under it.

### DIFF
--- a/crates/core/src/merge/record_merger.rs
+++ b/crates/core/src/merge/record_merger.rs
@@ -459,4 +459,56 @@ mod tests {
             ]
         );
     }
+
+    /// Records what the append-only strategy does with a delete block.
+    ///
+    /// It concatenates the data batches and returns, so the delete batches are
+    /// never consulted: a deleted record is still present in the output. This
+    /// test exists to pin that, because it is reachable — the table config
+    /// derives `append_only` whenever meta fields are disabled or no ordering
+    /// field is set, including on merge-on-read tables whose log files carry
+    /// delete blocks.
+    #[test]
+    fn test_append_only_does_not_apply_deletes() {
+        let schema = create_test_schema(false);
+
+        let data = RecordBatch::try_new(
+            schema.clone(),
+            vec![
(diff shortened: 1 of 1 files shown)


linliu-code:
Closing: already contained in the surviving line.

`test_append_only_does_not_apply_deletes` is the one difference between the old
#639-#662 tip and the squashed commit the current line starts from, so it is
present in #673 today (crates/core/src/merge/record_merger.rs). Keeping this
open would put the same test in two open PRs.

Review it in #673, or in #660 for the whole line squashed.

PR #657, opened 2026-08-03
test(core): read the merge-on-read fixtures against their Spark snapshots

linliu-code:
**Stacked on #639–#655** — their commits appear here until they merge. **Review only the last commit.**

## What

19 merge-on-read layout fixtures, the gold comparison helper, and a test that reads each fixture and checks the result against the `SELECT *` snapshot Spark dumped **inside** it. So this asserts agreement with Hudi's reference reader, not with this reader's own prior output.

The file slice is discovered from the extracted fixture rather than written down per case — a fixture that gains a file cannot silently stop being covered.

## Result: 4 of 15 read correctly

| | |
|---|---|
| **Pass** | `log_only`, `log_compaction`, `parquet_log_block`, `partial_update` |
| **Fail** | 11, each recorded with what stops it |
| **No gold** | 2 (`delete_ord_int`, `hfile_log_block`) |

**Every failure is in this crate's Avro-to-Arrow decoding, not in the merge:**

```
Decimal128(20, 2) not supported            ×2
Decimal128(30, 15) not supported           ×1
dictionary types other than UTF-8          ×1
Invalid magic / log format version         ×3
union index out of bounds / negative len   ×4
```

The first four are outright rejections; the rest are that decoder losing framing on a payload it mis-reads. **`parquet_log_block` passes**, which places the gap in the Avro path specifically rather than in log-block handling.

## This corrects something I claimed earlier

The port was landed on the premise that this crate's own `avro_to_arrow` could stand in for the forked `arrow-avro` upstream uses. It can for **schema conversion** — that is all the schema handler needed, and it works. It **cannot yet for decoding log block data**. That is why upstream moved to `arrow-avro`, and these fixtures are what surfaced it.

The gap list is an `#[ignore]`d test rather than a comment, so a gap that starts passing **fails the assertion** and gets promoted into the passing set rather than going unnoticed. Run `cargo test -- --ignored --nocapture` to see it.

## One fix included

The resolver now reads **`hoodie.record.merge.mode`**. A v9 table states its merge semantics in that key, and this crate reads it nowhere — it infers them from whether an ordering field is set. All 19 fixtures are `COMMIT_TIME_ORDERING` with **no ordering field**, so every one was inferred as `append_only` and refused. Same inference as #656, which shows it also silently drops deletes on the existing read path.

Full workspace suite green (1105 passing, 1 ignored); build warning-free.


linliu-code:

PR #659, opened 2026-08-03
feat(core): add a switch between the two merge implementations

linliu-code:
**Stacked on #639–#658** — their commits appear here until they merge. **Review only the last commit.**

## What

`hoodie.read.merge.engine = legacy | v2`, defaulting to `legacy`, dispatched inside the file group reader. **Nothing changes for anyone** — the default is the reader that has always served reads.

## Why land it now, before parity

The ported reader has been unreferenced across sixteen changes. This is the first one that shows **how it is meant to be reached, and from where** — and the first that reads anything through it from the public API.

```
read_file_slice_from_paths(..)
  ├─ is_metadata_table()        → existing reader   (always, regardless of setting)
  ├─ no log files / read-opt.   → base-file read    (nothing to merge; see below)
  ├─ engine == legacy           → existing reader   (default)
  └─ engine == v2               → adapter → ported reader
```

One seam. `Table`, DataFusion, the Python binding and the cxx bridge are untouched. The streaming entry point reaches it too, since a merge-on-read stream already collects through this method.

## What wiring the two together turned up

The switch is small; making it actually *read* was not. Three ways the ported reader's inputs did not match what this crate stores:

| | Symptom | Cause |
|---|---|---|
| **Properties escaping** | schema silently never resolved | `hoodie.table.create.schema` is stored as Java writes a properties file, with `:` escaped — not valid JSON unescaped. `sanitize_avro_schema_str` already existed for this; the new code wasn't using it. |
| **Meta fields** | `Key field '_hoodie_record_key' not found in schema` | The create schema is the schema the *user* wrote. The files carry the `_hoodie_*` columns and the merge needs the record key. |
| **End timestamp** | `Config(NotFound("hoodie.read.end.timestamp"))` | The context resolver requires one; this reader treats an absent bound as unbounded. The adapter now supplies the far-future instant rather than refusing reads the existing reader serves. |

Each was only reachable by actually performing a read — which is why the new test does one.

## Design points worth review

**A present-but-unparseable schema is an error, not a `None`.** Those are different situations, and reporting a broken table as "no schema was resolved" sends the reader looking in the wrong place.

**Base-file-only slices stay on the existing path.** Not a capability limit — the engine serves them fine, reducing to a base file read. But it takes its schema from the create schema where that path takes it from the base file, and the Avro conversion models a map as `Dictionary(Utf8, V)`: not a valid Arrow type (dictionary keys must be integers), and it doesn't reconcile against the `Map` the file really has. **Every parquet fixture in this repo has a map column**, so routing those reads through it would turn working reads into failures for the sake of uniformity. Deriving its schema from the base file instead would remove the difference — a change on its own, and one that would also close the standalone-entry schema gap.

**Filters and projection are not in the adapter.** Both engines' results run through the same `apply_eager_options`, so they cannot drift on what a filter means or which columns come back.

**Two refusals rather than fallbacks:** a metadata table always uses the existing reader (HFile, which the ported reader has no support for — permanent, see #658); a read with no resolvable schema is refused rather than guessed.

**An unparseable engine name errors.** `get_or_default` would have silently swallowed a typo — leaving a caller convinced they were exercising one engine while actually reading with the other. That is the one outcome a switch like this must not produce.

## Tests

Seven. The load-bearing one is the last:

- the default is `legacy`
- a metadata table stays on `legacy` even when `v2` is asked for
- asking for `v2` selects it
- an unknown value errors
- asking for `v2` does not change a base-file-only read
- asking for `v2` does not defeat read-optimized mode
- **`v2` actually reads a slice with log files, through the dispatch, and returns the same row count as the existing reader** — selecting an engine and reaching it are different things, and a dispatch wired to nothing would satisfy all six of the others

That last one is also the first differential assertion between the two engines. It holds here because this fixture's merge semantics agree; a general differential harness still waits on the merge-mode question.

## Not yet

`v2` stays opt-in. It is not at parity — the gap list lives in the fixture tests (#657). This PR is the plumbing and the decision points, not the cutover.

Full workspace suite green (1116 lib + 79 table-read + 39 datafusion); build warning-free.


linliu-code:

Issue #76, opened 2024-07-19
Support CoW CDC incremental query


xushiyan:

jonathanc-n:
@xushiyan Hallo, would like to work on this. High level overview:
- Just adding `LogFiles` to record changes made to data set. 
- 
Let me know if there is anything else you want accompanying this change?

xushiyan:
block on #9

PR #660, opened 2026-08-03
feat(core): serve merge-on-read reads with file group reader version 2

linliu-code:
## Why

[#677](https://github.com/apache/hudi-rs/pull/677) added `hoodie.read.file.group.reader.version` and defaulted it to `2`, but nothing was wired in behind the switch: `version_two_unsupported_reason` returned a reason for every read, so version 1 served all of them. This is the reader that switch was built for.

## What version 2 claims

Version 2 now serves every read except the ones it explicitly refuses. That inversion is the substance of this PR — the capability check went from "refuse unless enumerated" to "serve unless enumerated" — so the refusals are the contract worth reading:

| Situation | Behaviour | Why |
|---|---|---|
| Metadata table | version 1, decided before the check | Its base files and log blocks are HFile. Permanent, not transitional. |
| `CUSTOM` record merge mode, **and a merge actually happens** | **error** | Falling back would merge with version 1's derivation, which drops deletes on a commit-time-ordered table. Wrong rows are worse than a refusal. A copy-on-write or read-optimized slice consults no merger, so it is served normally. |
| Non-parquet base file | version 1 | Version 2 resolves the format from config alone; version 1 also falls back to the file extension. A Lance table that never sets `hoodie.table.base.file.format` is the real case. |
| `hoodie.datasource.write.drop.partition.columns=true` | version 1 | Neither reader reconstructs a dropped partition column from the partition path. Version 1 fails the projection; version 2 would treat it as a column a later writer added and null-fill it — a wrong value where there was a refusal. (`Table` rejects this configuration outright unless config validation is skipped.) |

Every refusal falls back rather than failing, and each is logged, so selecting a version cannot turn a working read into a failing one. `hoodie.read.file.group.reader.version=1` restores the previous reader exactly.

## Evidence

Each fixture is read twice — once per reader version — and compared against a snapshot Spark produced from the same table, cell by cell.

- **62 fixtures** compared against Hudi's own output; the known-disagreement list is a ratchet, so a new disagreement fails the build and a fixed one must be removed.
- **183 option cases** under version 2 across 23 fixtures — projection, read-optimized and incremental windows, generated from the same manifest the Spark snapshots were, so a case cannot be spelled differently on the two sides.
- **62 fixtures under position-based merge**, with `merging_by_position_and_by_key_disagree_on_duplicate_record_keys` proving the two strategies are distinguishable on the one fixture that can tell them apart — otherwise the sweep would pass if position merge silently fell back to key merge.
- Every remaining entry on the known-disagreement list names **version 1**. Version 2 has none.

## The four tests that would have to fail if this were wrong

Mutation-checked rather than assumed — a test that cannot fail proves nothing:

| Test | Injected fault | Result |
|---|---|---|
| `test_read_file_slice_from_paths_default_version_matches_version_one` | drop a row in `read_via_v2` | fails |
| `test_read_file_slice_from_paths_incremental_excludes_out_of_window_rows` | remove `apply_commit_time_filter` | fails |
| `test_streaming_agrees_with_eager_on_both_reader_versions` | drop a row in `read_via_v2` | fails |
| `test_decode_avro_is_partial_false_resolves_to_the_reader_schema` | read `IS_PARTIAL` by presence | fails |

The second replaces a test that passed with the filter deleted — [#671](https://github.com/apache/hudi-rs/pull/671) caught that, and it now runs against a compacted base file holding a pre-window row no log record replaces.

## What review removed

Review of this PR shrank it by ~1,200 lines. Worth stating, because a diff that
gets smaller under review usually means something was found:

- **A superseded spill codec.** `BufferedRecord::to_spill_bytes` /
  `from_spill_bytes` and their ordering-value encoding documented themselves as
  "used by `SpillableRecordMap` to store an entry in the on-disk tier". They are
  not: the map spills whole Arrow batches through `row_serde` and keeps ordering
  values in its in-memory index. A parallel implementation left behind by its own
  replacement, with a doc asserting the opposite — plus eight tests covering it.
- **`ProjectingBatchReader`**, constructed nowhere.
- **`reader_v2/gold_tests.rs`** (335 lines): six fixtures against Spark, with all
  ten of its `KNOWN_GAPS` entries stale. `tests/gold_parity_tests.rs` covers the
  same fixtures and 56 more, with a ratchet that fails when an entry goes stale.
- Fields never read and constructors never called: `InputSplit::{start,length}`
  (always `0`/`-1`), `DeleteRecord::partition_path` (always empty),
  `RecordPayload::into_record`, `BufferedRecords::from_engine_record`.

Every file-scope `#![allow(dead_code)]` in `reader_v2` is gone. What remains is
48 targeted allows, each naming why that item is kept — mostly Java-parity
surface reached only by the test harness, which drives the engine the way an FFI
caller does. Per-item rather than per-file, so the argument about any one of them
can be had on its own.

Internal tracker ids and phase labels (103 `ENG-*` references, `A1`/`A2`/`B2`,
and pointers to an internal design doc) are replaced by the behaviour they stood
for.

**One thing the sweep surfaced rather than removed:** `allow_inflight_instants`
is plumbed from the loader into a field nothing reads, so the knob is inert.
Nothing sets it true today, so behaviour is correct — but it is annotated saying
so rather than deleted, because deleting it would hide that the setting does
nothing.

## Behaviour changes

Three, all intentional:

1. **Incremental windows bound completion time** on timeline layout v2 (Hudi 1.x parity); layout v1 has no completion times and still bounds requested times. The translation happens in exactly one place, `Table::resolve_incremental_window` — translating twice selects nothing. Verified against Hudi's own gold for four boundary windows, including one that opens on a commit's requested time and must therefore keep that commit.
2. **A missing base file reports a schema-read failure** rather than an open failure, because version 2 resolves the schema first. Both still name the file and report not-found.
3. **`FileSlice::base_file` is now `Option<BaseFile>`**, and `base_file_relative_path()` returns `Result<Option<String>>`, so a log-only file group is representable. This is a breaking change to hudi-core's public Rust API for crates.io consumers upgrading to 0.5; in-repo consumers (datafusion, python, cpp) are updated.

## Reviewing this

The diff is far past what anyone can read, and pretending otherwise helps nobody. The load-bearing parts are small:

- `crates/core/src/file_group/reader.rs` — `serve_with_version_two` and `version_two_unsupported_reason`. One gate, shared by the eager and streaming paths so they cannot drift.
- The four tests above.
- `crates/core/tests/gold_parity_tests.rs` — the harness and its known-disagreement list.

Everything else is backed by the corpus. This PR replaces the ten stacked PRs #663–#673, whose review comments are all addressed here; their commits are the readable split of this change.

## Not covered

- A timestamp nested in a struct, list or map keeps arrow-avro's `+00:00` zone spelling rather than `UTC`. No fixture in the corpus has one, so it is documented as not yet implemented rather than implemented untested.
- Per-slice reader-selection lines are `debug!` ([#672](https://github.com/apache/hudi-rs/pull/672)), so diagnosing which reader served a table needs debug logging.


crates/core/src/file_group/reader_v2/input_split.rs
```diff
@@ -0,0 +1,259 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Ported from the merge-on-read reader. Nothing consumes it yet, so its
+//! items are unreachable from the crate's call graph until the reader wires in.
+#![allow(dead_code)]
+
+//! Mirrors `org.apache.hudi.common.table.read.InputSplit`.
```
hudi-agent, line 20: 🤖 nit: the `#![allow(dead_code)]` attribute is sandwiched between two `//!` module-doc blocks, splitting what reads as a single description across an attribute. Could you move `#![allow(dead_code)]` before the first `//!` line (or after the last), so the module doc is contiguous?

crates/core/src/file_group/reader_v2/iterator_mode.rs
```diff
@@ -0,0 +1,57 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Ported from the merge-on-read reader. Nothing consumes it yet, so its
+//! items are unreachable from the crate's call graph until the reader wires in.
+#![allow(dead_code)]
+
+//! Mirrors `org.apache.hudi.common.table.read.IteratorMode`.
```
hudi-agent, line 20: 🤖 nit: same split `//!` block issue as `input_split.rs` - the `#![allow(dead_code)]` attribute breaks the module doc in two. Could you move it so the description is contiguous?

crates/core/src/file_group/reader_v2/iterator_mode.rs
```diff
@@ -0,0 +1,57 @@
...
+    EngineRecord,
+
+    /// Yield HoodieRecord wrappers (not yet implemented in Rust).
+    HoodieRecord,
+
+    /// Yield only record keys (not yet implemented in Rust).
+    RecordKey,
+}
+
+impl IteratorMode {
```
hudi-agent: 🤖 nit: `from_str_opt` is a non-idiomatic name - the `_opt` suffix for returning `Option` is not a Rust convention. Have you considered implementing `std::str::FromStr` (enabling `.parse::<IteratorMode>()`) or naming it `try_from_str` to align with how the ecosystem spells fallible conversions?

crates/core/src/file_group/reader_v2/buffered_record.rs
```diff
@@ -0,0 +1,1094 @@
...
+                    None => {
+                        return Err(CoreError::ReadFileSliceError(
+                            "spill decode: null element in composite ordering value".to_string(),
+                        ));
+                    }
+                }
+            }
+            Ok(Some(OrderingValue::Composite(elems)))
+        }
+        other => Err(CoreError::ReadFileSliceError(format!(
```
hudi-agent: 🤖 nit: `is_empty` and `is_delete` are identical — could you remove `is_empty` and let callers use `is_delete` directly? The name `is_empty` implies something subtly different (a record with no meaningful content might not be a delete in a richer model), and having two names for the same predicate creates a maintenance surface where they could drift apart.

hudi-agent (commented):

test. A couple of naming and structural nits, particularly around the split module doc blocks and a non-idiomatic parse helper. a couple of naming/design nits below — `is_empty` aliasing `is_delete` and the Java-style unit-struct factory. a couple of small naming and doc-comment nits below. a couple of naming nits below. a couple of minor naming and doc-comment nits below; the logic and structure are clean. code looks clean overall — the new files are well-documented, the two-phase `DeleteContext` construction is clearly explained, and the gap-reporting module is a nice design for surfacing silent divergences. No style or readability issues worth flagging.

crates/core/src/table/mod.rs
```diff
@@ -155,6 +158,19 @@ impl Clone for Table {
     }
 }
 
+/// One tick below `instant_time`, as a same-width instant string.
+///
+/// Used to turn an exclusive start bound into an inclusive one. Instant times are
+/// fixed-width numeric strings, so this is a plain decrement; a value that does
+/// not parse (the bootstrap sentinels, for instance) is returned unchanged, which
+/// keeps the bound no narrower than it was.
+fn instant_time_minus_one(instant_time: &str) -> String {
```
yihua, line 178: This decrements the instant as an integer, so it can land on a string that sorts correctly but isn't a valid time — `20250713010500000 - 1` is `20250713010499999`, i.e. `SS=99`. That value becomes `StartTimestamp`, and `actions_in_range` parses it, so an incremental read whose earliest admitted commit has `ss=00` and `SSS=000` fails with `Timeline("Invalid epoch millis")` (I reproduced this by renaming a fixture's first commit). Could the exclusive bound be kept as a flag on the selector instead of being encoded as a "one tick below" string?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

.github/workflows/ci.yml
```diff
@@ -81,6 +83,29 @@ jobs:
...
+  # is the reason the feature exists — it needs no libclang and no C++ toolchain,
+  # so a consumer that cannot take a native build (or does not want one in a
+  # Python wheel) has a configuration to use. Nothing else in CI compiles it, so
+  # without this job it would break silently.
+  rust-tests-no-spill-backend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Cache dependencies
+        uses: swatinem/rust-cache@6323deb1c3ea99e6f4dae6b7d3ac1b73ac21ff77 # v2.9.2
```
yihua: This SHA doesn't resolve — `6323deb1c3ea99e6f4dae6b7d3ac1b73ac21ff77` isn't a commit in swatinem/rust-cache. v2.9.2 is `6323deb102c322ba6fcbdcafc7e3dddab59af2b6`, which is what the other two jobs in this file pin. The job currently fails at "Set up job", so the no-spill leg never runs. (The build itself is fine — I ran both commands locally.)

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/gold_tests.rs
```diff
@@ -0,0 +1,335 @@
...
+/// single field is the primitive, so `IntWrapper` and a bare `long` encode to
+/// the same bytes. Only the branch index differs. Reading both therefore needs
+/// to key off the writer version rather than the payload, which is a design
+/// question rather than a decoding one — and swapping to the newer schema
+/// outright breaks the fixtures written the older way.
+///
+/// **A corrupt tail block is not recognised.** `LogFileReader` has no
+/// corrupt-block detection — `create_corrupted_block_if_needed` returns `None`
+/// — so it parses the trailing garbage as a block.
+const KNOWN_GAPS: &[(&str, &str)] = &[
```
yihua: Running this with `--ignored` fails its own second assertion: all ten `KNOWN_GAPS` entries now read correctly, including the four `table_delete_ord_*` ones. The doc above ("two causes remain — avro maps as Dictionary, wrapped delete ordering values") is stale in the same way. Since `tests/gold_parity_tests.rs` covers the same fixtures with a live ratchet, is this harness still earning its keep, or could it be folded in?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Context types for the merge-on-read file group reader.
+//!
+//! This module holds the inputs the MOR reader needs and the resolver that
+//! derives them from a table's configs. Nothing consumes it yet — the reader
```
yihua: This doc and the file-scope `allow(dead_code)`s were written for the stacked PRs, but this is the change that wires the reader in, so they're now hiding real dead code in a live module. Dropping the 15 allows surfaces 38 warnings — `HoodieFileGroupReaderBuilder` and `ProjectingBatchReader` are never constructed, `RecordContext::partition_path` is threaded through three layers and never read, and the spill codec's `to_spill_bytes`/`from_spill_bytes` are unused. Worth clearing before merge while it's still cheap to tell intentional from accidental?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader.rs
```diff
@@ -264,9 +430,13 @@ impl FileGroupReader {
     /// selecting a version cannot turn a working read into a failing one. Each
     /// reason is logged, because a fallback nobody can observe is
     /// indistinguishable from a reader that is never used.
+    ///
+    /// Refusals are the exception, not the fallthrough: version 2 is the reader
```
yihua, line 442: The paragraph just above ("Every reason here means version 1 serves the read instead, so selecting a version cannot turn a working read into a failing one") is contradicted by the CUSTOM branch below, which returns `Unsupported`. Since version 2 is now the default, a MOR table with `hoodie.record.merge.mode=CUSTOM` that reads today will start failing. Could the doc carve that case out, and the error message name `hoodie.read.file.group.reader.version=1` as the way back?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/buffer/spillable_map.rs
```diff
@@ -0,0 +1,2627 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Ported from the merge-on-read reader. Nothing consumes it yet, so its
+//! items are unreachable from the crate's call graph until the reader wires in.
+#![allow(dead_code)]
+
+//! Size-tracked, RocksDB-spillable merge map (A1, ENG-42993).
```
yihua: There are a few dozen `ENG-XXXXX` ids and phase labels (A1, A2, A3c, A6e, B2, I-3, I-33, G-16) left in the source — e.g. this line, `record_context.rs:1790`, `read_stats.rs:72`. They don't resolve for anyone reading this on the ASF side. Could they be replaced with the behaviour they're standing in for, or a hudi-rs issue number?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/Cargo.toml
```diff
@@ -101,12 +107,32 @@ percent-encoding = { workspace = true }
...
+# Also a regular dependency, but only an optional one behind `spill-rocksdb`.
+# Repeated here because tests outside the spill tier use it too, and must keep
+# building in the `--no-default-features` configuration.
 tempfile = { workspace = true }
 
 [lints.clippy]
 result_large_err = "allow"
 
 [features]
+default = ["spill-rocksdb"]
```
yihua, line 119: The feature's rationale is that the Python wheel and the cxx bridge shouldn't have to build rocksdb, but it's default-on and `crates/hudi` takes `hudi-core` with default features, so both still do. Should the umbrella crate forward the feature so those two can opt out, or is the intent that they keep it and only external consumers opt out?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/tests/reader_concurrency_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+        "{label:34} reader version={reader_version} rows before={before} after={after}  {}",
+        if after == before {
+            "OK (orphan excluded)"
+        } else {
+            "*** UNCOMMITTED ROWS VISIBLE ***"
+        }
+    );
+}
+
+#[ignore = "documents a divergence from Java: layout-v1 tables do not filter uncommitted files"]
```
yihua: Running this with `--ignored` prints "OK (orphan excluded)" for all six combinations, so the reason on the `#[ignore]` no longer reproduces — which the module doc above says should be updated when it happens. Did the divergence get fixed somewhere in this change, or does it need a different fixture to show up?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/record_merger.rs
```diff
@@ -0,0 +1,799 @@
...
+            }
+        }
+    }
+
+    /// Mirrors Java `EventTimeRecordMerger.finalMerge`.
+    ///
+    /// The newer (log) record wins unless it is NOT a commit-time-ordering delete
+    /// and the older record (also not such a delete) has a strictly greater,
+    /// same-typed ordering value.
+    fn final_merge(
```
yihua, line 169: Java's `finalMerge` delegates straight to `shouldKeepNewerRecord`; this re-derives the same rule inline. They agree today (I worked through the CTO-delete and cross-type cases), but it's one rule in two places. Any reason not to call `should_keep_newer_record(older_record, newer_record)` here?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

yihua (commented):
Thanks for working on this — wiring version 2 in behind the switch from #677, and the gold corpus that backs it, is a lot of real ground covered, and `tests/gold_parity_tests.rs` is genuinely good (its coverage assertions and staleness ratchet mean it can't quietly stop testing things, and I confirmed every entry on its known list names version 1). Two things I'd want fixed before merge: `instant_time_minus_one` (`table/mod.rs:167`) decrements an instant as an integer, so it can produce a calendar-invalid string that `actions_in_range` then rejects — an incremental read whose earliest admitted commit has `ss=00` and `SSS=000` fails with `Timeline("Invalid epoch millis")`, which I reproduced by renaming a fixture's first commit (no fixture in the corpus has that shape, which is why the sweep is green); and the new `rust-tests-no-spill-backend` job pins a `swatinem/rust-cache` SHA that doesn't exist, so it fails at "Set up job" and the leg it guards has never run (that leg passes locally, so it's the guard that's broken). The rest of the inline comments are about the module having outgrown its scaffolding — `gold_tests.rs`'s `KNOWN_GAPS` list is entirely stale (all ten now pass, and its doc still describes delete-ordering limitations that no longer hold), `reader_v2/mod.rs` still says nothing consumes the reader while its file-scope `allow(dead_code)`s hide 38 real dead items, the CUSTOM-merge-mode doc contradicts the branch directly above it, and there are several dozen `ENG-XXXXX` ids and internal phase labels in the source that AGENTS.md rules out for this repo. On process rather than code: `max(added, deleted)` here is ~45k against the repo's 1000-line guideline, and I don't think any one reviewer can honestly claim to have read it — the stacked #663–#673 were the reviewable form of this, and I'd rather land those than sign off on the squash; two more points belong in the description rather than the diff, namely that the `hoodie.read.start/end.timestamp` completion-time change applies regardless of reader version (so `=1` isn't an escape hatch for it) and that the Python docstrings (`python/src/internal.rs:283`, `_internal.pyi:389`) still describe the old semantics. This is one review pass over the load-bearing parts you flagged plus the timeline and config changes — not an approval, and not a claim to have read the corpus code.

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/table/mod.rs
linliu-code: Confirmed and fixed — thank you, this was a real bug and the reproduction was exact. I reproduced it before changing anything:

```
20250713010500000 -> 20250713010499999   parse: ERR Timeline("Invalid epoch millis")
20250713010501000 -> 20250713010500999   parse: OK 2025-07-13 01:05:00.999 UTC
```

The `ss` field becomes 99. It sorts correctly, so every range comparison looked right, and it is not a time, so the selector rejects it.

**I tried your suggested fix first — carrying the exclusivity as a flag on the selector — and it broke `hudi_table_get_file_slices_incremental` (2 slices instead of 3).** `StartTimestamp` has four consumers building ranges, not one: `create_instant_range_for_log_file_scan`, `resolver::resolve_instant_range`, the row-mask path, and the metadata-table reader. Making only the log-scan range inclusive dropped a slice from the file-slice selection path, which never goes through it. So the flag approach is right in principle but needs all four moved together, and I did not want to land a partial version of it inside this PR.

What I did instead removes the invalid value rather than the encoding: the decrement now steps in the **time** domain, so the result is both a real instant and lexicographically just below the input.

```rust
const FORMATS: [&str; 2] = ["%Y%m%d%H%M%S%3f", "%Y%m%d%H%M%S"];
if let Ok(dt) = Instant::parse_datetime(instant_time, "UTC") {
    for format in FORMATS {
        if dt.format(format).to_string() == instant_time
            && let Some(stepped) = dt.checked_sub_signed(TimeDelta::milliseconds(1))
        { ... }
    }
}
```

The round-trip is what distinguishes a date-formatted instant from an epoch-millis one (a metadata table's `00000000000000000`) — both parse, only the former renders back to itself — so the sentinels keep the integer decrement.

Two tests, both mutation-checked against the old integer decrement:

- `test_instant_time_minus_one_stays_a_valid_instant` — every borrow boundary: whole second, whole minute, whole hour, midnight (the date steps back), second-precision instants, and the epoch-millis sentinels.
- `test_resolve_incremental_window_start_bound_is_a_real_instant` — the bound sorts below the earliest admitted instant **and** parses. No fixture has a commit on a whole second, which is why this went unseen, so it asserts the property rather than reproducing the one input that broke.

I'd still like the flag version eventually — it removes the dual contract instead of satisfying it — but as its own change, with all four consumers moved at once.

.github/workflows/ci.yml
linliu-code: Fixed — you and I found this within about an hour of each other; it was pushed just before your comment landed.

Confirmed the same way you did: `gh api repos/Swatinem/rust-cache/commits/6323deb1c3ea99e6f4dae6b7d3ac1b73ac21ff77` returns `422 No commit found`. It shares the `6323deb1` prefix with the real v2.9.2 SHA and carries the same `# v2.9.2` comment, so it reads as correct at a glance — and the job died in "Set up job" after two seconds, before checkout, so nothing of ours ever ran.

Now pinned to `6323deb102c322ba6fcbdcafc7e3dddab59af2b6`, matching the other two jobs. I also checked every pinned action SHA across all workflow files resolves against the API — `astral-sh/setup-uv` and `docker/setup-buildx-action` both do. The `rust-tests-no-spill-backend` leg passes on the current head.

crates/core/src/file_group/reader.rs
linliu-code: You're right, and it was my paragraph. Fixed both halves.

The doc now carves the case out explicitly rather than claiming something the code below contradicts:

> Every reason here **but one** means version 1 serves the read instead… The exception is a `CUSTOM` record merge mode with a merge to perform, which errors. That one *does* turn a working read into a failing one: version 2 is the default, so a merge-on-read table declaring `CUSTOM` that reads today stops reading. It is deliberate — falling back would merge with version 1's own derivation, which drops deletes, and wrong rows are worse than a refusal — but it is a behaviour change, and the error names the way back.

And the error now names the escape hatch:

> A table with a CUSTOM record merge mode needs its own merger, which no reader here implements. Set `hoodie.read.file.group.reader.version=1` to read it with the reader that served it before, which merges without that merger.

Worth recording alongside this: the only merge-on-read tables in the fixture corpus with `record.merge.mode=CUSTOM` are the metadata tables, which route to version 1 before the capability check runs. So this refusal is covered by unit tests only — no gold fixture exercises it.

crates/core/src/file_group/reader_v2/mod.rs
linliu-code: Confirmed, and I've made a start rather than a sweep.

You're right that the framing is stale: 25 files still say "nothing consumes it yet", written when these were stacked PRs, and this is the change that wires the reader in. I removed the file-scope `allow(dead_code)` from `input_split.rs` and `iterator_mode.rs` (their module docs needed unsplitting anyway — see the two nits on those files), and it immediately surfaced exactly what you describe: `InputSplit::start` and `InputSplit::length`, always `0` and `-1`, never read. Both removed.

I have **not** done the other 13 allows. Clearing them properly means deciding per item whether something is scaffolding the port still needs or an accident, and I would rather that be a deliberate pass — with your read on `HoodieFileGroupReaderBuilder` and `ProjectingBatchReader` in particular, since "never constructed" could mean either. Happy to do it in this PR if you'd prefer it cleared before merge; my instinct is a follow-up so the diff stays reviewable, but it's your call and I don't feel strongly.

crates/core/src/file_group/reader_v2/gold_tests.rs
linliu-code: Confirmed — running it with `--ignored` fails exactly as you say:

```
these now read correctly and should move to GOLD_FIXTURES:
["table_column_projection", "table_all_data_types", "table_null_containers",
 "table_corrupt_tail_block", "table_delete_ord_int", "table_delete_ord_long",
 "table_delete_ord_double", "table_delete_ord_decimal",
 "table_delete_ord_string", "table_delete_ord_timestamp"]
```

All ten, including the four `table_delete_ord_*` you named, and the "two causes remain" doc above is stale with them.

On whether it earns its keep: I think you're right that it doesn't. `tests/gold_parity_tests.rs` covers the same fixtures against the same Spark snapshots with a live ratchet — a new disagreement fails the build and a fixed one has to be removed — whereas this harness is `#[ignore]`d, so its data can rot without anything noticing, which is precisely what happened. A second copy that cannot fail is worse than no copy.

I've left it alone for now rather than deleting it unilaterally, since removing a test harness on someone else's read is not my call to make silently. Say the word and I'll fold what it uniquely covers into `gold_parity_tests.rs` and delete it.

crates/core/tests/reader_concurrency_tests.rs
linliu-code: Confirmed — `--ignored` gives 3 passed, 0 failed, and the orphan-exclusion probe prints "OK (orphan excluded)" for all six combinations. The reason on the `#[ignore]` no longer reproduces, and the module doc above says that is exactly when it should be updated.

I don't yet know which of your two possibilities it is — fixed somewhere in this change, or needing a different fixture to show up. The honest answer is that I would have to bisect the stack to tell, and I did not want to guess in the ignore reason. Left as-is pending that; if you'd rather it not merge in this state I'll dig into which commit changed the behaviour and either drop the test or re-point it at a fixture that still shows the divergence.

crates/core/Cargo.toml
linliu-code: Your reading is right, and it makes the feature's stated rationale not currently true for the two consumers it names. `spill-rocksdb` is default-on, `crates/hudi` takes `hudi-core` with default features, and both the Python wheel and the cxx bridge go through the umbrella crate — so both still build rocksdb, libclang and all.

The intent was the second of your two options — external consumers opt out — but the doc reads as though the wheel and the bridge already benefit, which they do not. Two ways to make it honest: forward the feature from `crates/hudi` so those two can actually opt out, or correct the rationale to say only external consumers benefit today.

I'd rather not pick between those inside this PR — forwarding the feature changes the umbrella crate's public feature surface, which deserves its own diff and its own note. Which would you prefer?

crates/core/src/file_group/reader_v2/buffer/spillable_map.rs
linliu-code: Confirmed and agreed: 105 `ENG-XXXXX` references across 16 files, plus the phase labels. They are internal tracker ids and mean nothing to a reader on the ASF side — the project's own guidance says not to reference internal plans.

I have not done this yet. It is mechanical in shape but not by `sed`: each one is standing in for a behaviour, and replacing it well means reading the surrounding code and writing down what the id was pointing at — otherwise the comment loses information rather than gaining it. Doing that badly across 105 sites would be worse than leaving them.

So: happy to do it, and I think it should be done before this merges into an ASF repo, but as a focused pass. Would you rather it land in this PR or immediately after?

crates/core/src/file_group/reader_v2/record_merger.rs
linliu-code: No good reason — you're right that it is one rule in two places, and thank you for working through the CTO-delete and cross-type cases to confirm they agree today.

I have not changed it yet, only because `final_merge` is on the merge hot path and I want to confirm the delegation is behaviour-identical under the gold sweep rather than by reading, given the two spellings agree by inspection but the whole point of the finding is that they could stop agreeing. That check is cheap and I'll do it — I just did not want to slip a hot-path change in unverified alongside everything else in this round.

If you'd rather see it in this PR, say so and I'll make the change and run the corpus against it.

crates/core/src/file_group/reader_v2/input_split.rs
linliu-code: Fixed. The `#![allow(dead_code)]` is gone rather than moved — see the discussion on `reader_v2/mod.rs`: this is the change that wires the reader in, so "nothing consumes it yet" was no longer true, and dropping the attribute surfaced two genuinely dead fields (`InputSplit::start` and `length`, always `0`/`-1`, never read) which are now removed. The module doc is a single contiguous block.

crates/core/src/file_group/reader_v2/iterator_mode.rs
linliu-code: Fixed the same way as `input_split.rs`: the attribute is removed rather than relocated, and the module doc is contiguous.

crates/core/src/file_group/reader_v2/iterator_mode.rs
linliu-code: Fixed — implemented `std::str::FromStr`, so `"ENGINE_RECORD".parse::<IteratorMode>()` works and the error is a `ConfigError::InvalidValue` naming the value rather than a bare `None`.

Worth noting it had no callers at all, so the idiomatic spelling cost nothing.

crates/core/src/file_group/reader_v2/buffered_record.rs
linliu-code: Fixed — `is_empty` removed, callers use `is_delete`. Your reasoning is the right one: they had identical bodies, `is_empty` implies something subtly different from "is a delete tombstone", and two names for one predicate is a drift surface. The three call sites were all in tests.

crates/core/src/file_group/reader_v2/mod.rs
linliu-code: Done — swept, and it found more than a lint.

Every file-scope `#![allow(dead_code)]` in `reader_v2` is gone (23 of them). Deleted outright:

- **The per-record spill codec.** `to_spill_bytes` / `from_spill_bytes`, `encode_`/`decode_ordering_value` and nine `SPILL_*` tags. Its doc claimed *"Used by `SpillableRecordMap` to store an entry in the on-disk (RocksDB) tier"* — it is not. `spillable_map` spills whole Arrow batches through `row_serde::to_binary_row_body` and keeps ordering values in the in-memory `DiskEntry` index. A superseded parallel implementation whose doc asserted the opposite. Eight tests went with it, including three I had added the round before — they were covering dead code.
- `ProjectingBatchReader`, constructed nowhere in the lib or under `cfg(test)`.
- `InputSplit::{start,length}` (always `0`/`-1`), `DeleteRecord::partition_path` (every construction site passed `String::new()`), `RecordPayload::into_record`, `BufferedRecords::from_engine_record`.

On your two specifically: `ProjectingBatchReader` is gone; `HoodieFileGroupReaderBuilder` turned out to be **used by the test harness**, which is why I checked rather than deleted — my first attempt removed it and broke the harness build.

What remains is 48 targeted allows, each naming why that item is kept, so the argument about any one of them can be had per item rather than per file. Mostly Java-parity surface reached only by the harness.

One thing worth flagging separately: `allow_inflight_instants` is plumbed from the loader into a field nothing reads, so the knob is **inert**. Nothing sets it true, so behaviour is correct today — I annotated it saying so rather than deleting it, since deleting would hide that the setting does nothing. Happy to make the builder refuse `true` instead if you'd prefer it loud.

`make check-rust` clean with zero dead-code warnings; 1267 lib tests green.

crates/core/src/file_group/reader_v2/gold_tests.rs
linliu-code: Deleted, agreeing with your read.

All three of its tests were subsumed: `merged_reads_match_the_spark_snapshot` covered 6 fixtures where `tests/gold_parity_tests.rs` covers 62 with a live ratchet; `known_gaps_still_fail` was `#[ignore]`d with all ten entries stale; and `fixtures_without_gold_are_present` is the same assertion `EXPECTED_WITHOUT_GOLD` already makes.

Your framing was the deciding argument — a second copy that cannot fail is worse than no copy, and its rotting data is exactly what that predicts.

crates/core/src/file_group/reader_v2/buffer/spillable_map.rs
linliu-code: Done — zero `ENG-*` references left, plus the phase labels (`A1`, `A2`, `B2`) and three pointers to an internal design doc.

103 references, but only **15 distinct ids**, so each one meant a single thing and could be replaced consistently with the behaviour it stood for: `ENG-42992` → the lazy base-file source, `ENG-42993` → the spillable merge map, `ENG-43009` → the vectorized merge path, `ENG-38318` → event-time ordering, and so on. Where the id was a bare parenthetical the prose already stood on its own, so it just goes.

Worth admitting how it went: my first pass carried a whitespace-normalising rule that destroyed indentation across all 15 files — code bodies, Apache license headers, and doc-comment list continuations. `rustfmt` recovered the code, and I restored the rest by diffing each file against the previous commit and taking the original wherever only whitespace differed. It is clean and verified now, but it was self-inflicted and I would not script prose edits that way again.

crates/core/src/file_group/reader_v2/record_merger.rs
linliu-code: Done — `final_merge` now delegates, matching Java's `finalMerge`.

I worked the case analysis first rather than trusting that they looked the same. `final_merge` returned the older record iff `!ctod(newer) && !ctod(older) && same_class && old > new`; `should_keep_newer_record` returns false under exactly that condition — the `new >= old` negation lines up. So:

```rust
if should_keep_newer_record(older_record, newer_record) {
    Ok(newer_record.clone())
} else {
    Ok(older_record.clone())
}
```

The gold corpus is green behind it: 62 fixtures and 183 option cases against Spark's output, plus the position-merge sweep. Thanks for confirming the CTO-delete and cross-type cases — that made this a rename rather than an investigation.

crates/core/tests/reader_concurrency_tests.rs
linliu-code: Bisected, and the answer is the first of your two possibilities: **it got fixed, in the very commit that added the probe.**

`cd0d5be` ("fix(core): address the merge-on-read port and concurrency reviews") says so itself:

> A file was readable whenever a completion timestamp could be found for its commit, which asked the wrong question twice. Timeline layout v1 records no completion timestamps at all, so nothing was filtered there and a snapshot returned rows from commits that never completed — 4 rows became 7 on the v6 fixtures.

That commit replaced the layout-v2-only check with `completed_requests: HashSet<String>`, populated for **both** layouts, plus the archival boundary. The probe was written as the reproduction, the fix landed alongside it, and the `#[ignore]` reason was never updated — which is why it now prints "OK (orphan excluded)" for all six combinations.

So it is an ordinary test now, exactly as this file's own module doc prescribes: `test_a_base_file_from_an_uncommitted_instant_is_not_readable`, un-ignored, asserting `after == before` instead of printing it, on both timeline layouts and both reader versions.

I also corrected the module doc, which claimed the `#[ignore]`s exist for two reasons (unfixed divergence, or non-asserting probe). Only the second applies now — there are no divergence ignores left in this file.

crates/core/schemas/HoodieDeleteRecord.avsc
yihua, line 4: Why is this `HoodieDeleteRecord` different than `HoodieDeleteRecordList.HoodieDeleteRecord`?

crates/core/src/file_group/reader_v2/resolver.rs
```diff
@@ -0,0 +1,908 @@
...
+        has_bootstrap_base_file: false,
+        needs_bootstrap_merge: false,
+        enable_logical_timestamp_field_repair: false,
+        // Predicate pushdown into the merge path has no caller here, so no
+        // filter is installed and the primary-key-safety gate is irrelevant.
+        row_filter_builder: None,
+        mor_pk_safe: false,
+        // The table-version < 8 completion gate needs a timeline the caller
+        // has not loaded; leaving it unset keeps the gate a no-op.
+        completion_gate_inputs: None,
```
yihua, line 89: This is the only production producer of `completion_gate_inputs`, so Gate 3 in `forward_scan_pass1` is inert for every read through this crate — the straddling case it exists for (writer A inflight at T1, writer B commits T2 > T1, so A's uncommitted blocks pass Gate 2) still merges. Not a regression — version 1 has no committed-set check either — but `Table` already holds the timeline this needs. Is wiring it through a follow-up, or deliberately off?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/schema/delete.rs
```diff
@@ -37,9 +37,54 @@ static DELETE_RECORD_AVRO_SCHEMA_IN_JSON: Lazy<Result<JsonValue>> = Lazy::new(||
...
+            "Expected a record for delete record".to_string(),
+        ));
+    };
+    let Some((_, ordering_val)) = fields.get_mut(2) else {
+        return Err(CoreError::Schema(
+            "Delete record has no orderingVal field".to_string(),
+        ));
+    };
+    if let AvroValue::Union(pos, inner) = ordering_val {
+        if *pos == ARRAY_WRAPPER_POSITION {
```
yihua, line 70: `ArrayWrapper` is what multiple ordering fields serialize to (HUDI-9569), and composite ordering is otherwise supported here — so a table with `hoodie.table.ordering.fields=ts,seq` reads fine until its first delete block, then the whole read errors. Could this degrade to natural order, as other unusable ordering values do? All six `table_delete_ord_*` fixtures are single-field, so the sweep can't see it.

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/engine.rs
```diff
@@ -0,0 +1,2077 @@
...
+
+        // Multi-field (composite) precombine/ordering keys ARE supported.
+        // `RecordContext::new` splits a comma-separated `hoodie.table.precombine.field`
+        // / `hoodie.table.ordering.fields` into `ordering_field_names`, and
+        // `get_ordering_values` builds one `OrderingValue::Composite` per row from
+        // the per-field scalars (compared lexicographically field-by-field, mirroring
+        // Java `OrderingValues`). A field absent from a batch, an unsupported field
+        // type, or a null component falls back to natural order — matching the scalar
+        // path — so there is no silent first-field-only degradation. (Construction
+        // still rejects composite *virtual keys* above: that path reconstructs the
```
yihua: This closing sentence — "Construction still rejects composite *virtual keys* above" — contradicts both the start of this same comment and `test_composite_virtual_keys_accepted_at_construction` below; the rejection it refers to no longer exists.

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/buffer/position_based.rs
```diff
@@ -0,0 +1,1491 @@
...
+        // tombstone never spilled (hudi-rs delete blocks always carry record
+        // keys, and `process_next_deleted_record` preserves them in the
+        // in-memory tier). A delete without a usable record key cannot be
+        // re-keyed — Java keeps it position-keyed under the hybrid strategy, but
+        // that requires the position, which `drain_iter` does not surface; fail
+        // loudly rather than silently drop a delete (which would resurrect a
+        // deleted row). A SPILLED position-keyed delete comes back with its
+        // position string as `record_key` and is indistinguishable from a real
+        // numeric key — persisting the record key through the spill tier is the
+        // real fix (tracked as a follow-up outside this buffer).
```
yihua: This looks already fixed: `record_from_entry` in `spillable_map.rs` persists and restores a delete's real key via `DiskLoc::Delete { record_key }`, with a comment naming this exact hazard — and I couldn't reproduce the mis-key described here. Could the comment be updated so the next reader doesn't chase a fixed bug?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/reader_context.rs
```diff
@@ -0,0 +1,445 @@
...
+    pub schema_handler: FileGroupReaderSchemaHandler,
+    pub table_config: HashMap<String, String>,
+    /// Per-read overrides. Populated by
+    /// [`resolve_reader_context`](crate::file_group::reader_v2::resolver::resolve_reader_context)
+    /// from the `hoodie.read.*` keys plus the individually-named ones its
+    /// `READER_CONFIG_KEYS` lists — anything a consumer looks up here must
+    /// appear in one of those two, or it reads as unset.
+    pub hoodie_reader_config: HashMap<String, String>,
+    /// Optional parquet `RowFilter` builder for predicate pushdown into base
+    /// parquet files and parquet-format log blocks. Set by the FFI bridge from
```
yihua: There are ~95 comments across `reader_v2/` referencing infrastructure that isn't in this repo — "the FFI bridge", "decoded substrait predicate", "Velox consumes the stream directly", `cpp/src/predicate.rs` (cpp/src has only lib.rs and util.rs), "#76/#95 review". For `row_filter_builder` and `completion_gate_inputs` it's more than tidiness: their only documented producer doesn't exist here, so inert-by-design and inert-by-accident read the same. Could these be rewritten against this crate's actual call graph?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/log_record_reader.rs
```diff
@@ -0,0 +1,1805 @@
...
+        self.progress = 0.0;
+        self.total_log_files = 0;
+        self.total_log_blocks = 0;
+        self.total_log_records = 0;
+        self.total_corrupt_blocks = 0;
+        self.total_rollbacks = 0;
+
+        let timezone = self.reader_context.timezone();
+
+        // Read all blocks as metadata-only (no content decoding).
```
yihua: This comment describes the lazy sweep, but the loop below uses the eager whole-file reader — the lazy tier (`new_streaming`, `read_all_blocks_metadata_only`, `LogBlockFetcher`/`DeferredContent`, and the `hoodie.memory.dfs.buffer.max.size` window knob) has no production caller, only the `#[cfg(test)]` memory_bench; the one production `load_content` call in Pass 3 no-ops because content is always already read. Since the range here is deliberately unbounded (rollback command blocks must always process), eager decode means out-of-window and rolled-back blocks pay a full Arrow decode before the gates discard them — v1 skips those at the header. Is wiring the sweep in a follow-up, or should the tier and the knob wait until it lands?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/log_file/content.rs
```diff
@@ -17,35 +17,128 @@
...
+
+    let mut active: Option<i8> = None;
+    for i in 0..union.len() {
+        let type_id = union.type_id(i);
+        match active {
+            None => active = Some(type_id),
+            Some(seen) if seen == type_id => {}
+            Some(seen) => {
+                return Err(CoreError::LogBlockError(format!(
+                    "Delete block mixes ordering types (union branches {seen} and {type_id})"
```
yihua, line 94: Null is a union branch too, so a delete block over a nullable ordering column where some deleted rows carry null mixes branch 0 with a wrapper branch and fails here. Java resolves the branch per record and treats a null as the default ordering, so the delete wins naturally. Not a regression — the old per-record narrowing couldn't read this shape either — but the message reads as if two value types mixed, when null+value is the realistic case. Could the null branch be exempted from the mixing check and those slots treated as default-ordering deletes?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/file_slice.rs
```diff
@@ -29,7 +29,20 @@ use std::path::PathBuf;
 /// a [FileSlice] is a logical group of [BaseFile] and [LogFile]s.
 #[derive(Clone, Debug)]
 pub struct FileSlice {
-    pub base_file: BaseFile,
+    /// The base file, when the slice has one.
+    ///
+    /// A slice written by inserts that went straight to log files has none —
+    /// Flink ingestion, a bucket index's first write to a bucket, and any
+    /// merge-on-read file group before its first compaction all produce these.
+    pub base_file: Option<BaseFile>,
```
yihua, line 37: `base_file` becoming `Option<BaseFile>` (and `base_file_relative_path()` returning `Result<Option<String>>`) is a breaking change to hudi-core's public Rust API — the field is `pub` and external code compiles against the old shapes directly. All in-repo consumers (datafusion, python, cpp) are updated and CI is green, so this is only about communication. Could the PR description call it out alongside the timestamp-semantics breaking change, for crates.io consumers upgrading to 0.5?

_⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality._

crates/core/src/file_group/reader_v2/resolver.rs
```diff
@@ -0,0 +1,908 @@
...
+
+    // Resolved by `Table::prepare_reader_options` for table-level reads. A
+    // `FileGroupReader` built straight from a base URI never loads the timeline,
+    // so the caller must supply it; defaulting here would silently widen the read.
+    let latest_commit_time: String = hudi_configs
+        .try_get(HudiReadConfig::EndTimestamp)?
+        .ok_or_else(|| ConfigError::NotFound(HudiReadConfig::EndTimestamp.as_ref().to_string()))?
+        .into();
+
+    let merge_mode = resolve_merge_mode(hudi_configs)?;
```
yihua: **Blocking:** this resolves the merge mode unconditionally, so a base-file-only read (COW or read-optimized) of a table declaring `CUSTOM` errors here even though `version_two_unsupported_reason` deliberately serves it — reproduced on a COW fixture with `hoodie.record.merge.mode=CUSTOM`, which fails every eager read under the default while the streaming path (returning early before the gate) serves it fine. Could this skip `resolve_merge_mode` when `has_log_files` is false, with an end-to-end test for the no-merge CUSTOM case? Also, the payload-class-inferred CUSTOM error doesn't name the `version=1` way back the way the gate's error does.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/reader_v2/buffer/key_based.rs
```diff
@@ -0,0 +1,7459 @@
...
+            };
+            // Single-probe merge (perf): probe `key` once via `merge_in_place` and
+            // overwrite in place, instead of get(probe+clone) → delta_merge → insert
+            // (probe+clone+probe). The merger only reads `existing` by reference. A2
+            // semantics preserved: the merged BatchRef payload is stored in-memory
+            // (IPC serialization deferred to spill only).
+            let merger = &self.base.buffered_record_merger;
+            self.base
+                .records
+                .merge_in_place(key, |existing| merger.delta_merge(&record, existing))?;
```
yihua, line 1491: **Non-blocking (follow-up):** when the existing record wins this delta merge (event-time, higher ordering), the prior is stored unchanged, so a stale partial's unique columns are dropped — Java folds the loser in both directions (`DefaultSparkRecordMerger.partialMerge` calls `mergePartialRecords(newer, older)` when older wins). E.g. C1 partial-updates `a` at ts=9, then C2 partial-updates `b` at ts=2: Hudi keeps C2's `b`, this keeps the base's. The IGNORE_DEFAULTS/FILL_UNAVAILABLE branch below does fill both directions, so it's just this narrow-schema path; a fixture with out-of-order partials would pin it.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/log_file/content.rs
```diff
@@ -17,35 +17,128 @@
...
+                )));
+            }
+        }
+    }
+    let Some(active) = active else {
+        return Ok(ordering.clone());
+    };
+
+    let child = union.child(active);
+    // Null is a branch like any other; there is nothing to unwrap out of it.
```
yihua: **Non-blocking (follow-up):** collapsing the union to one column makes two Java-legal delete blocks fail the read — an all-null `orderingVal` decodes to a `DataType::Null` column that `delete_batch_to_keys_with_ordering` rejects loudly, and a block mixing wrapper types (e.g. one null among longs) errors above as "mixes ordering types". Java decodes each DeleteRecord independently and merges both fine, and a delete written without a precombine value is the most default-shaped delete there is. Worth a follow-up plus fixtures for both.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/reader_v2/resolver.rs
yihua, line 89: **Non-blocking (question):** `completion_gate_inputs` is `None` at every wired construction site, so the inflight/completion gate never runs and a multi-writer inflight append to an already-sliced log file reads as committed. Version 1 has the same gap so this isn't a regression, but the Pass-1 docs read as if the gate were wired — is populating this from `Table`'s timeline a planned follow-up?

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/reader_v2/log_record_reader.rs
```diff
@@ -0,0 +1,1805 @@
...
+        let unbounded_range =
+            InstantRange::new(self.reader_context.timezone(), None, None, false, true);
+        profile_once!(self.log_block_read_ms, {
+            for path in &self.log_file_paths.clone() {
+                // A5 / R5: stream the log file in bounded windows for the Pass-1
+                // header sweep instead of buffering the whole file. Each block
+                // gets a ranged fetcher so Pass-3 inflate reads only its content.
+                // Upstream sweeps block headers in bounded windows and inflates
+                // each block's content later. This crate's log file reader is
+                // whole-file, so the blocks come back with their content already
```
yihua: **Non-blocking (follow-up):** the wired scan decodes every block's content before the gates run, so a decode failure inside a block the gates would discard (e.g. a rolled-back instant's block) fails the read, and a whole slice's decoded batches sit in memory before merging (the merge map spills; this vec doesn't). The streaming sweep that fixes both exists but is only called from `memory_bench.rs` — is wiring it a planned follow-up?

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/tests/gold_parity_tests.rs
```diff
@@ -0,0 +1,986 @@
...
+    SweepResult {
+        reported,
+        entry_used,
+        without_gold,
+        with_options,
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn every_fixture_matches_hudi_on_both_reader_versions() {
```
yihua, line 855: **Non-blocking (test gap):** no fixture in the corpus exercises event-time "stale loses" — neither a lower-precombine log update nor a stale delete — so a regression flipping version 2 to last-writer-wins under EVENT_TIME_ORDERING would pass this entire sweep. That semantics currently lives only in unit/buffer tests; one fixture with base ts=100 and a later log upsert at ts=50 would close the corpus's biggest blind spot.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

yihua (commented):
Thanks for working on this! I did a correctness pass focused on the merge semantics (inserts/updates/deletes) and the incremental window handling, checked against Java's `BufferedRecordMergerFactory`/`DefaultSparkRecordMerger` and validated empirically (full `hudi-core` suite on the PR head: 1398 passed / 0 failed, gold sweep included). One inline comment is blocking — the CUSTOM merge-mode resolver error that breaks no-merge reads, reproduced end-to-end — and the rest are non-blocking follow-ups: partial-update delta-merge parity, null/mixed-wrapper delete blocks, the unwired completion gate, eager block decode, an event-time coverage gap, and a docs sweep. This is one review pass, not an approval.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/reader_v2/log_record_reader.rs
```diff
@@ -0,0 +1,1805 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Ported from the merge-on-read reader. Nothing consumes it yet, so its
```
yihua: **Non-blocking (nit):** this module doc still says "Nothing consumes it yet / unreachable from the crate's call graph", but this PR wires it into the production path — same stale line in `merged_log_record_reader.rs`, `buffer/loader.rs`, `record_merger.rs`, `delete_context.rs`, `update_processor.rs`, and `record_buffer.rs`. Worth a sweep so the docs don't contradict the wiring.

⚠️ AI-generated; verify before applying. React 👍/👎 to flag quality.

crates/core/src/file_group/reader_v2/resolver.rs
yihua: Fixed: the resolver now tolerates an unsupported merge mode when there are no log files to merge (commit-time ordering stands in for the mode nothing reads), the merging-read error names `hoodie.read.file.group.reader.version=1`, and end-to-end tests pin the COW, read-optimized, and merging cases.

crates/core/src/schema/delete.rs
yihua: Fixed: `unwrap_ordering_values` decodes ArrayWrapper rows as null ordering cells, so those deletes apply in natural order (warn-logged) instead of failing the read, with a regression test.

crates/core/src/file_group/log_file/content.rs
yihua: Fixed: the null branch is exempt from the mixing check and decodes as null cells that merge as natural-order deletes, matching Hudi, with regression tests for the mixed and all-null shapes.

crates/core/src/file_group/log_file/content.rs
yihua: Addressed with the thread above: null and ArrayWrapper branches decode per row as natural-order deletes, and only a genuine two-scalar-type mix still errors.

crates/core/src/file_group/file_slice.rs
yihua: Done: the PR description's behaviour-changes section now calls out the `FileSlice::base_file` / `base_file_relative_path()` API break for crates.io consumers upgrading to 0.5.

crates/core/src/schema/delete.rs
yihua: Correction: rather than degrading, the union column now passes through whole and `delete_batch_to_keys_with_ordering` decodes ArrayWrapper per row into the composite ordering value the merge compares, so a stale composite delete loses correctly, with regression tests for the decode and the merge outcome.

crates/core/src/file_group/log_file/content.rs
yihua: Correction to the above: ArrayWrapper rows now decode to composite ordering values (union passthrough) rather than natural order; null branches still decode as natural-order deletes.

yihua (approved):
LGTM overall as the implementation of the new file group reader - core merge semantics look correct; deletes are handled properly with ordering values; MOR log file reading covers log block ordering, corrupt block, rollback block, and partial updates; schema evolution is supported; incremental queries are supported on both table v6 (requested time-based) and v8+ (completion time-based).  Non-blocking review comments can be addressed in follow-ups.

yihua:
The lines that miss coverage are codecov artifact: tarpaulin's llvm engine misses lines in async bodies that tests demonstrably execute. The merge semantics are validated scenario-by-scenario against Spark-written reference test fixture tables on both reader versions.  Merging this PR.

crates/core/schemas/HoodieDeleteRecord.avsc
yihua: Answering my own question after tracing it: the list file is the wire schema (wrapper-record union, what Java serializes) and this file is its post-unwrap companion, where each union position holds the bare primitive inside the wrapper at the same position, which `avro_schema_for_delete_record` indexes to build the narrowed reader schema after `unwrap_ordering_value` strips the wrapper (pinned by `test_schema_for_delete_record`). Since only their own tests still call those helpers, a follow-up could either cross-reference the two files in a comment or drop this one with the helpers.

PR #663, opened 2026-08-04
fix(core): read the delete record's wrapped ordering value

linliu-code:
**Stacked on #639–#662** — review only the last commit.

## The bug

A delete block whose ordering value is anything but a small integer fails to read:

```
Union index 1490 out of bounds: 2
Cannot convert i64 to usize: -26
```

"out of bounds: **2**" is the tell — the only 2-branch unions are `recordKey` and `partitionPath`. So the decoder is reading a *record key's* union index and getting 1490. The byte stream is misaligned; the branch wasn't chosen wrongly.

**This is a pre-existing hudi-rs defect, not a port regression.** It fails on the current reader too, through `Table::read`.

## Cause

`orderingVal` is declared here as a union of primitives. Hudi writes a union of per-type **wrapper records**, and inserted `BooleanWrapper` at position 1 — so every position from `int` onward names a different type:

| Position | This crate | Hudi (writer) |
|---|---|---|
| 1 | int | BooleanWrapper |
| 2 | long | IntWrapper |
| **3** | **float** | **LongWrapper** |
| 4 | double | FloatWrapper |
| … | … | … |

## Evidence

The delete block in `table_delete_ord_long`, 26 bytes, is exactly self-consistent under Hudi's schema:

```
04 | 02 02 34 | 02 00 | 06 c0 3e | 02 02 33 | 02 00 | 06 f0 2e | 00
```

- `04` → 2 records
- keys `"4"` and `"3"`, empty partitions, ordering position **3**, varints **4000** and **3000**

Decoded here, position 3 is a `float` — 4 fixed bytes — so `c0 3e 02 02` is swallowed and `33` is read as the next record's key union index: zigzag(0x33) = **−26**, the reported error, byte for byte.

## What changed

1. **The schema** — Hudi's wrapper union, for decoding.
2. **The narrowing template** — position *i* now holds the primitive that wrapper *i* carries, so the existing "keep `[null, i]`" logic still yields a primitive Arrow column and the output type is unchanged.
3. **Unwrapping** — `Union(i, Record[("value", v)])` → `Union(1, v)`, applied **after** narrowing, since narrowing reads the position Hudi wrote and unwrapping rewrites it.
4. **A cast** — the wrapper is chosen for the *value*, not the column, so a long ordering column can carry an `IntWrapper` for a small value. The delete batch's ordering column is cast to the type the data schema declares. The old schema hid this by calling position 2 a long regardless.
5. **`ArrayWrapper` is rejected** in both places that read the position, rather than mapped to something that disagrees with its value.

## A correction to an earlier claim

I previously reported that both shapes exist in the wild and could not be distinguished, after an attempt to swap the schema broke `v6_trips_8i3d` and `v8_trips_8i3u1d`. That was wrong. Both use position **2**, which is `long` here and `IntWrapper{int}` there — and a wrapper's field is a bare primitive, so **both decode the same bytes to the same number**. Those fixtures never distinguished the schemas. They broke on the Arrow *type* of the result, which is what item 4 addresses.

All 16 fixtures with delete blocks are consistent with the wrapper schema; none requires the primitive form.

## Older tables

Not supported, deliberately. No fixture uses the primitive layout, and the two cannot be told apart from the bytes: where they differ, decoding usually fails (this bug), and at position 2 both succeed with the same value. **If you know of tables written with the older layout, please say so** — the alternative is a decode-and-retry fallback, which I did not take.

## Tests

11 previously-ignored cases un-ignored and passing: ordering by long, decimal, timestamp, plus multi-log, watermark and instant-range cases that were blocked behind the same decode.

The schema's own unit tests asserted the old index space; they now assert the wire order, which is what a delete block's index actually refers to. Two new tests cover unwrapping and the `ArrayWrapper` rejection.

Full workspace green: 1177 lib + 79 table-read + 39 datafusion + 21 + 12. Ignored drops 19 → 8.


crates/core/src/file_group/reader_v2/engine.rs
```diff
@@ -0,0 +1,1978 @@
...
+        // records, then stripped by the buffer before output. The column is NOT
+        // added to `required_schema`/`merge_schema` — only to the base source's
+        // physical schema (`base_read_schema` = required + row-index).
+        let use_position = self.use_record_position();
+        // Unused: position-based merging is not wired up here.
+        let _row_number_col = use_position.then(|| ROW_INDEX_TEMPORARY_COLUMN_NAME.to_string());
+        let base_read_schema: SchemaRef = if use_position {
+            let mut fields: Vec<arrow_schema::FieldRef> =
+                required_schema.fields().iter().cloned().collect();
+            fields.push(Arc::new(arrow_schema::Field::new(
```
hudi-agent, line 848: 🤖 When `use_record_position` is true, `base_read_schema` gains a non-nullable `ROW_INDEX_TEMPORARY_COLUMN_NAME`, but the base read only projects the file∩required intersection, so that column is never present in `raw`. `project_batch_to_schema` (line 894, and the streaming map at ~930) then hits its non-nullable-absent branch and errors the whole read. Since the comment below says the row number is "dropped rather than faked," did you mean to leave the column out of `base_read_schema` too (or make it nullable) until position merge is wired up?

crates/core/src/avro_to_arrow/arrow_array_reader.rs
```diff
@@ -849,6 +879,30 @@ fn resolve_u8(v: &Value) -> Option<u8> {
...
+/// scale is applied by the array's type, not here. The bytes are narrower than
+/// 16 whenever the value fits in fewer, so they are sign-extended.
+///
+/// Returns `None` for a value that is not a decimal, or whose unscaled value is
+/// too wide for `i128`; the row then reads as null rather than as a wrong number.
+fn resolve_decimal128(v: &Value) -> Option<i128> {
+    let bytes: &[u8] = match v {
+        Value::Bytes(b) => b,
+        Value::Fixed(_, b) => b,
+        _ => return None,
```
hudi-agent, line 895: 🤖 `resolve_decimal128` handles `Value::Bytes`/`Value::Fixed`, but `from_avro_datum` decodes a decimal-logical-type field to `Value::Decimal` — and since the `Decimal128` arm is only reached for `Schema::Decimal` fields, every real decimal value falls to `_ => return None` and reads as null. Could you add a `Value::Decimal(d)` arm (its big-endian bytes come from `Vec::<u8>::try_from(d)`)? The same nulling hits delete records ordered by `DecimalWrapper` (bytes + logicalType decimal).

crates/core/src/file_group/reader_v2/input_split.rs
```diff
@@ -0,0 +1,258 @@
...
+        paths.sort_by(|a, b| {
+            let name_a = a.rsplit('/').next().unwrap_or(a);
+            let name_b = b.rsplit('/').next().unwrap_or(b);
+            match (LogFile::from_str(name_a), LogFile::from_str(name_b)) {
+                (Ok(lf_a), Ok(lf_b)) => lf_a.cmp(&lf_b),
+                _ => a.cmp(b), // fallback to lexicographic if parsing fails
+            }
+        });
+        paths
+    }
```
hudi-agent, line 121: 🤖 nit: the doc comment says "no base file and no log files" but the implementation only checks `!self.has_log_files()` — a base-only split returns `true` here too. Could you correct the doc (e.g. "Returns true when there are no log files to merge") and maybe rename to `is_base_only` or `needs_no_merge` to match?

crates/core/src/file_group/reader_v2/update_processor.rs
```diff
@@ -0,0 +1,190 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//! Ported from the merge-on-read reader. Nothing consumes it yet, so its
+//! items are unreachable from the crate's call graph until the reader wires in.
+#![allow(dead_code)]
+
+//! Mirrors `org.apache.hudi.common.table.read.UpdateProcessor`.
```
hudi-agent, line 24: 🤖 nit: having two inner `//!` doc blocks in a row is a bit odd — the second one (starting at line 25) overwrites the first in rustdoc. Could you merge them into a single `//!` block at the top, keeping both the porting note and the Java mirror description?

crates/core/src/file_group/reader_v2/update_processor.rs
```diff
@@ -0,0 +1,190 @@
...
+            self.num_deletes.fetch_add(1, Ordering::Relaxed);
+            return Ok(None);
+        }
+        // handleNonDeletes: prev present → update; prev absent → insert.
+        if previous_record.is_some() {
+            self.num_updates.fetch_add(1, Ordering::Relaxed);
+        } else {
+            self.num_inserts.fetch_add(1, Ordering::Relaxed);
+        }
+        Ok(Some(merged_record.clone()))
```
hudi-agent, line 127: 🤖 nit: the `_emit_deletes` parameter is always `false` in practice (as the doc note explains), so it currently has no effect on the returned type. Could you rename it to `emit_deletes` and add a `todo!()` or `debug_assert!(!emit_deletes)` to make the unimplemented branch explicitly visible, rather than silently ignoring it?

crates/core/src/file_group/reader_v2/memory_limit_tests.rs
```diff
@@ -0,0 +1,291 @@
...
+const MERGE_TYPE_PAYLOAD_COMBINE: &str = "payload_combine";
+
+const MIB: u64 = 1024 * 1024;
+
+/// Threshold `SpillConfig` derives from a given `hoodie_reader_config` map.
+fn threshold_bytes(config: &HashMap<String, String>) -> u64 {
+    SpillConfig::from_config(config)
+        .expect("SpillConfig::from_config")
+        .max_in_memory_size
+}
```
hudi-agent, line 72: 🤖 nit: could you add the `test_` prefix here (and to the other `oom_*` functions below)? The project convention is `test_<function>_<scenario>_<expected>`, and names like `i33_merge_budget_honored_when_forwarded` / `oom_peak_cap_rejects_oversized_insertion_loudly` don't show up clearly when filtering test output with `--test-filter test_`.

hudi-agent (commented):

Thanks for working on this! This PR fixes reading a delete record's ordering value when it's stored as a Hudi per-type wrapper (e.g. `LongWrapper`) by adopting Hudi's wrapper-union schema and unwrapping after narrowing, and it lands a large reader_v2 merge-on-read port (buffered records, log-record reader, file-group reader/engine, Avro→Arrow decimal/timestamp handling, and test scaffolding), currently `#![allow(dead_code)]`-gated. Two spots are worth a second look: the position-merge base read schema in engine.rs and the decimal128 decode path in arrow_array_reader.rs, which appears to null out real values. Please take a look at the inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. One misleading doc comment in `input_split.rs`; the rest of the code is clean. a few small naming and doc nits below. a couple of naming and layout nits below, but the code overall is clean and well-documented. a few naming and simplification suggestions below — mostly test-name consistency and one opaque return type. a couple of minor readability suggestions below — `get_record_context` uses the Java-style `get_` prefix that's un-idiomatic in Rust, and `generate_required_schema` has six parameters that would read more clearly as a named options struct. a couple of naming nits below — one redundant/misleading method name, one constant whose "MAX" label reads backward. a few naming and clarity nits below.

crates/core/src/file_group/reader_v2/engine.rs
linliu-code: Checked this against the code and I don't think it reproduces — the row-index column is *virtual*, not projected.

`ParquetBaseFileReader` filters the row-index name out of the projection indices and emits the column anyway (`base_file/parquet.rs`, the `.filter(|name| Some(name.as_str()) != options.row_index_column.as_deref())` in the projection branch, with a comment saying exactly this). So `raw` does carry `ROW_INDEX_TEMPORARY_COLUMN_NAME` and `project_batch_to_schema` finds it rather than hitting the non-nullable-absent branch.

Three pieces of evidence:

1. `test_row_index_column_is_returned_alongside_a_projection` in `base_file/parquet.rs` pins the behaviour directly.
2. `engine.rs` has a test that drives `make_base_file_source(false)` — the exact eager path here — with `use_record_position: true`, and asserts the positions come back as `[0, 1, 2]`.
3. The gold sweep runs all 62 fixtures under position merge (`every_fixture_matches_hudi_when_merging_by_record_position`), several of which were written by Hudi with `RECORD_POSITIONS` headers, so the path executes rather than erroring.

No change made. Thanks for the close read — the shape of the concern was right, the projection is just not the whole story.

crates/core/src/avro_to_arrow/arrow_array_reader.rs
linliu-code: Already resolved later in this stack. `crates/core/src/avro_to_arrow/arrow_array_reader.rs` no longer exists in the squashed tree — delete-block decoding moved to arrow-avro, and `561b781 fix(core): decode Avro decimals, and carry the UTC zone on timestamps` addressed the decimal gap you're describing (the schema side produced `Decimal128` with no matching array-builder arm, so any log block with a decimal column failed the read).

The case you name specifically — delete records ordered by `DecimalWrapper` — is covered by the `table_delete_ord_decimal [MorAvro]` fixture, which ships a `gold_options` manifest and matches Hudi's own output across the full option matrix under both reader versions, with no entry on the known-disagreements list.

crates/core/src/file_group/reader_v2/update_processor.rs
linliu-code: Fixed. Renamed to `emit_deletes` and added a `debug_assert!(!emit_deletes, "emitting deletes is not implemented; buffer/loader.rs rejects it upstream")`, with the doc saying the `true` branch is not yet implemented. `todo!()` would have been wrong here — the parameter exists for signature parity with gold and the value is always `false`, so panicking on a value nothing produces would be dead code that only fires if the upstream rejection regresses. The `debug_assert` catches exactly that case without changing release behaviour.

crates/core/src/file_group/reader_v2/memory_limit_tests.rs
linliu-code: Fixed. All five now carry the `test_` prefix: `test_i33_merge_budget_honored_when_forwarded`, `test_oom_current_footprint_getter_reflects_in_memory_bytes`, `test_oom_peak_cap_rejects_oversized_insertion_loudly`, `test_oom_no_cap_preserves_unchanged_behavior`, `test_oom_peak_cap_config_parses_and_defaults_to_none`.

crates/core/src/file_group/reader_v2/input_split.rs
linliu-code: Fixed. Renamed to `is_base_only` and corrected the doc to "Returns true when there are no log files to merge, so the split reduces to its base file. A base-file-only split is exactly this case." The single call site in `engine.rs` is updated.

crates/core/src/file_group/reader_v2/update_processor.rs
linliu-code: Fixed. Merged into one `//!` block, with the Java-mirror description first and the porting note plus `#![allow(dead_code)]` after it, so nothing is overwritten in rustdoc.

linliu-code:

PR #665, opened 2026-08-04
fix(core): take the merge-on-read schema from the base file

linliu-code:
**Stacked on #639–#664** — review only the last commit.

## What was wrong

The merge-on-read reader took its schema from `hoodie.table.create.schema` — the schema the table was **created** with.

Hudi treats that as a last resort. `TableSchemaResolver` resolves in this order:

1. the latest completed commit's metadata (the schema that write actually used)
2. a base file's footer
3. `hoodie.table.create.schema`

And this crate already implements the same order in `schema::resolver::resolve_data_schema`. The reader was reaching past both better sources for the weakest one.

## What it does now

Reads the base file's own schema — Hudi's tier 2, and what the existing read path effectively uses, so **the two engines now start from the same types**. It is also what the data actually has: under schema evolution the create schema is stale, and the engine evolves each batch to the required schema regardless.

## Three workarounds removed

Each of these existed only to make the create schema usable:

| Workaround | Why it was needed |
|---|---|
| Unescaping `\:` | The value arrives as Java writes a properties file, and is not valid JSON until unescaped |
| Prepending `_hoodie_*` | The create schema is the user's schema; the files carry meta fields and the merge needs the record key |
| Refusing when absent | A table that never recorded one could not be read — **including every reader built from a bare base URI**, which is the shape the cxx bridge uses |

The third is the one that mattered: it was a standing gap for the standalone entry point, and it closes here without new plumbing, because every read already has a base file path.

## Base-file-only slices now route through the engine

They were held back in #659 because the create schema modelled an Avro map as `Dictionary(Utf8, V)` — an invalid Arrow type — and every parquet fixture here has a map column. The schema no longer comes from there, and the conversion itself is fixed in #664, so the guard is gone.

The engine reduces to a base file read for these. The test asserting that setting `v2` does not change such a read now genuinely **compares the two engines** rather than one path against itself.

## Cost

One footer request. The engine reads the footer again when it opens the file; collapsing the two is worth doing and is not this change.

## Tests

No new ones — this removes special-case handling rather than adding behavior, and the existing coverage tightens as a result: the base-file-only comparison becomes a real differential test, and every harness case that reads a slice now exercises the base-file schema path.

Full workspace green: 1180 lib + 79 table-read + 39 datafusion + 21 + 12.


crates/core/src/file_group/reader.rs
```diff
@@ -244,7 +329,17 @@ impl FileGroupReader {
             .collect();
         let base_file_only = log_file_paths.is_empty() || options.is_read_optimized()?;
 
-        let merged = if base_file_only {
+        // Read-optimized means the log files are not read at all, so hand the
+        // engine none rather than letting it merge them. A slice with no log
+        // files reduces to a base file read either way.
+        let merged = if self.merge_engine()? == MergeEngineValue::V2 {
```
hudi-agent, line 335: 🤖 For incremental reads, `read_base_file_eager` filters base rows per-row by `_hoodie_commit_time` (`apply_commit_time_filter`), but routing base-file-only slices through `read_via_v2` swaps that for the engine's file-level `apply_instant_range_filter`. For a CoW/read-optimized incremental read under `merge.engine=v2`, wouldn't a base file created inside the window still carry forward older rows and return them, where the legacy path filters them out? The snapshot-only equality test wouldn't catch this.

crates/core/src/file_group/reader.rs
```diff
@@ -199,6 +199,91 @@ impl FileGroupReader {
...
+    /// This is what the existing path effectively reads with, so the two engines
+    /// start from the same types. It is also what the data actually has: under
+    /// schema evolution `hoodie.table.create.schema` records the table as it was
+    /// created, and the engine evolves each batch to the required schema anyway.
+    ///
+    /// Reading the footer costs one request. The engine reads it again when it
+    /// opens the file; collapsing the two is worth doing but is not this change.
+    async fn resolved_data_schema(
+        &self,
+        base_file_path: &str,
```
hudi-agent, line 237: 🤖 nit: the new implementation never returns `Ok(None)` — it either produces a schema or an `Err` — so `Result<Option<SchemaRef>>` quietly misleads callers into guarding a `None` branch that can never fire. Could you simplify the return type to `Result<arrow_schema::SchemaRef>` and update the call site accordingly?

hudi-agent (commented):

Thanks for working on this! This PR switches the merge-on-read reader's up-front schema from `hoodie.table.create.schema` to the base file's own footer schema, drops the three create-schema workarounds, and routes base-file-only slices through the v2 engine. The schema change itself reads cleanly and the motivation is well documented. The main thing worth double-checking is in the inline comment: whether routing base-file-only slices through the v2 engine changes incremental-read semantics, since the legacy path applies a per-row commit-time filter that the v2 file-level instant-range filter does not. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. One naming/API nit worth addressing: the `Option` in the return type is now vestigial.

crates/core/src/file_group/reader.rs
linliu-code: This was a real gap when you raised it, and a later commit in the stack closed it: `read_via_v2` now ends with `apply_commit_time_filter(&self.hudi_configs, merged)`, so the per-row `_hoodie_commit_time` mask is applied to version 2's output too — after the merge rather than before, because a row's commit time is whichever record won.

You were also right that the snapshot-only equality test wouldn't catch it. I mutation-tested the guard: deleting that call is caught by `incremental_admits_a_commit_by_completion_time_not_requested_time` and by `version_two_drops_a_base_record_from_outside_the_window`, which reads a compacted base file where `d` is a pre-window row that no log record replaces.

Separately, the test that was *supposed* to guard this at the file-group level turned out to be vacuous — see my reply on #671, where it's been rewritten against a fixture that can actually fail.

crates/core/src/file_group/reader.rs
linliu-code: Fixed. `resolved_data_schema` now returns `Result<arrow_schema::SchemaRef>` — you're right that it either produced a schema or an `Err`, so the `Option` only invited callers to guard a branch that could never fire. Both call sites (`read_via_v2` and `stream_via_v2`) are updated.

linliu-code:

PR #667, opened 2026-08-06
feat(core): detect corrupt log blocks and resume at the next one

linliu-code:
**Stacked on #639–#665** — review only the last commit.

## What was there

```rust
fn create_corrupted_block_if_needed(&mut self, _pos: u64, _len: Option<u64>) -> Option<LogBlock> {
    // TODO: support creating corrupted block
    None
}
```

A corrupt or truncated block failed the entire log file. `BlockType::Corrupted` existed in the model but nothing could ever produce one.

## What it does now

A well-formed Hudi log block records its total size **twice** — in the header length field and again in a trailing reverse pointer — and is followed by either another block or EOF. Three checks:

1. the trailing pointer has to lie inside the file
2. the size it records has to agree with the header
3. what follows the block has to be a magic marker or the end

Every offset is computed with `checked_add`/`checked_sub`, so a garbage length **reports corruption rather than panicking or allocating against it**. That is why the check runs before the body is parsed rather than after — a bogus content length would otherwise reach a `vec![0u8; len]`.

When a block fails, the reader emits a corrupt marker and resumes at the next magic marker, found by scanning in 1 MB windows that overlap by `MAGIC.len() - 1` so a marker straddling a boundary is not missed. **One bad block now costs its own span instead of the rest of the file.**

Adopted from the internal reader, which has carried this and the recovery tests for some time.

## Tests

Two harness cases un-ignored and passing — a corrupt tail block, and a delete-ordering fixture that was failing behind the same read.

Three unit tests on the check itself, since the harness cases only exercise one shape:

- a length disagreeing with the trailing pointer is corrupt, and the real length is **not** — the negative case matters, since a check that always says "corrupt" would pass the positive one
- a length past EOF, and `u64::MAX`, are corrupt — decided arithmetically, never by reading there
- the recovery offset always lands within the file

Full workspace green: 1185 lib + 79 table-read + 39 datafusion + 21 + 12. Ignored 7 → 5.

## Note

While verifying this I hit 117 unrelated failures that turned out to be a **stale fixture-extraction cache** under `$TMPDIR/hudi-rs-test-fixtures`, not the change — they reproduce on a clean tree. Clearing the directory fixes it. The cache key includes the zip's mtime, so it should self-invalidate; worth a look separately if it recurs in CI.


crates/core/src/file_group/log_file/reader.rs
```diff
@@ -321,6 +491,64 @@ mod tests {
...
+        assert!(
+            reader.is_block_corrupted(0, 1 << 40)?,
+            "a length past EOF is corrupt"
+        );
+        Ok(())
+    }
+
+    /// Recovery lands on the next magic marker, or the end of the file when
+    /// there is none — so one bad block costs its own span, not the rest.
+    #[tokio::test]
```
hudi-agent, line 538: 🤖 nit: the test name says `finds_eof_when_no_further_magic`, but it drives `scan_for_next_block_offset(0)` against a valid multi-block avro file, so the function will land on the *next* valid magic rather than EOF — and the assertion (`offset <= len`) would pass either way. Could you either use a single-block fixture so the name matches what's exercised, or rename to something like `test_scan_for_next_block_offset_stays_within_file_bounds`?

crates/core/src/file_group/log_file/reader.rs
```diff
@@ -99,32 +126,164 @@ impl<R: Read + Seek> LogFileReader<R> {
...
+        let result = if corrupt {
+            true
+        } else {
+            self.reader
+                .seek(SeekFrom::Start(block_end))
+                .map_err(CoreError::ReadLogFileError)?;
+            !self.next_is_magic_or_eof()?
+        };
+
+        self.reader
```
hudi-agent, line 229: 🤖 nit: the doc says "at or after `from_pos`" but the implementation immediately advances by `MAGIC.len()` before scanning, so `from_pos` itself is never a candidate. Could you update the wording to something like "strictly after `from_pos`'s own magic" so callers don't expect the current magic to be returned?

hudi-agent (commented):

Thanks for working on this! This PR gives `LogFileReader` real corrupt-block detection: it validates each block's trailing reverse-pointer and block-end MAGIC (with checked arithmetic) before parsing the body, and on an anomaly synthesizes a `BlockType::Corrupted` block and resumes at the next MAGIC (or EOF), so one bad span no longer fails the whole file.

I traced the detection arithmetic against the actual on-disk bytes of the log fixtures and confirmed the `trailing == block_length + MAGIC.len()` invariant it depends on holds for the real Java-written files; the reader-position handling, the recovery-scan window overlap (verified exhaustively), termination/progress, and the downstream `Corrupted` handling in Pass 1 / Pass 3 / `detect_content_type` all check out. Since `StorageReader` buffers the whole object in memory, the extra per-block seeks are in-memory, not remote round-trips.

One out-of-scope heads-up for follow-up: `gold_tests.rs` still lists `("table_corrupt_tail_block", "no corrupt-block detection")` in `KNOWN_GAPS` and has a doc comment referencing the now-deleted `create_corrupted_block_if_needed`; those are now stale (the `known_gaps_still_fail` test is `#[ignore]`d, so normal CI is unaffected).

No correctness issues found. A few style/readability suggestions in the inline comments. Please take a look, and this should be ready for a Hudi committer or PMC member to take it from here. A couple of naming nits below — a misleading test name and a doc comment that understates where scanning begins.

cc @yihua

crates/core/src/file_group/log_file/reader.rs
linliu-code: Fixed, taking your second suggestion: renamed to `test_scan_for_next_block_offset_stays_within_file_bounds`. The doc now says outright that the fixture is a valid multi-block file so the scan lands on the *next* block's magic rather than EOF, and that what's pinned is that either answer stays inside the file — the old name claimed something this fixture cannot produce.

crates/core/src/file_group/log_file/reader.rs
linliu-code: Fixed. The doc now reads "Offset of the next MAGIC strictly after the magic at `from_pos`", and says why: the scan starts `MAGIC.len()` bytes in, so a caller recovering from a bad block at `from_pos` would otherwise be handed it straight back.

linliu-code:

PR #668, opened 2026-08-06
fix(core): decode Avro log blocks with arrow-avro

linliu-code:
**Stacked on #639–#667** — review only the last commit.

## The bug

A decimal column in an Avro log block reads as **entirely NULL**.

```rust
fn resolve_decimal128(v: &Value) -> Option<i128> {
    let bytes = match v {
        Value::Bytes(b) => b,
        Value::Fixed(_, b) => b,
        _ => return None,      // ← Value::Decimal lands here
    };
```

A declared `logicalType: decimal` decodes to `apache_avro::Value::Decimal`, not `Bytes` or `Fixed`. Confirmed by round-tripping the exact fixture schema (`fixed(9)`, precision 20, scale 2):

```
DECODED VARIANT = Decimal
```

`None` there is indistinguishable from "this field was genuinely null", so the column comes back empty rather than failing. Same silent-wrong-answer shape as the map bug in #664 — and from the same converter.

## Why replace rather than patch

The vendored `avro_to_arrow` (adopted from DataFusion) has now produced two silent wrong answers. `arrow-avro` handles both correctly out of the box, and is maintained upstream. Verified against the fixture that has both problem columns:

```
COL decimal_field       type=Decimal128(20, 2)  null_count=0/1
COL nullable_map_field  type=Map(...)           null_count=0/1
```

**No arrow upgrade** — `arrow-avro` is published at 57, matching the workspace.

## Two shapes reconciled

**Framing.** A Hudi data block frames each datum with a four-byte length and no Avro framing; `arrow-avro`'s decoder expects a Single Object Encoding prefix. The writer schema is registered once, and the ten bytes that yields — marker plus Rabin fingerprint — are written ahead of each body into a buffer reused across records.

**Timezone spelling.** `arrow-avro` writes a UTC zone as `+00:00`; parquet writes `UTC`. Same zone, but Arrow compares timezones as strings, so a log batch refuses to concatenate with the base batch it merges with:

```
It is not possible to concatenate arrays of different data types
  (Timestamp(µs, "UTC"), Timestamp(µs, "+00:00"))
```

Decoded batches are relabelled to `UTC`. The values are already UTC instants, so nothing is converted.

## Also: it is columnar

The replaced path built an `apache_avro::Value` per record — an enum allocating a `String` per string field and a `Vec` per collection — then walked that to fill the columns. `arrow-avro` decodes into Arrow buffers directly. On log-heavy merge-on-read reads that is the hot path.

## Scope

The vendored converter **stays for now**: delete blocks decode from values that are already parsed, a different shape that this decoder doesn't take. Moving that too is a follow-up, after which the converter can be deleted.

## Tests

Two previously-ignored harness cases un-ignored and passing (`harness_all_data_types`, `harness_mixed_column_types`). No new tests: this swaps an implementation, and the existing coverage — 19 gold fixtures compared against Spark's own output, plus 58 harness cases — is what proves it.

Full workspace green: 1186 lib + 79 table-read + 39 datafusion + 21 + 12. **Ignored 5 → 3**, and the remaining three are unrelated (schema promotion, an HFile expectation, the gold gap list). Clippy clean in the changed files.


crates/core/src/file_group/log_file/avro.rs
```diff
@@ -16,77 +16,148 @@
...
+        matches!(tz, "+00:00" | "+0000" | "00:00" | "Z" | "z")
+    }
 
-        // Create reader expecting 1 record but with insufficient data
-        let mut avro_reader = AvroDataBlockContentReader::new(reader, &schema, 1);
+    let needs_fix = batch
+        .schema()
+        .fields()
+        .iter()
+        .any(|f| matches!(f.data_type(), DataType::Timestamp(_, Some(tz)) if is_utc_alias(tz)));
```
hudi-agent, line 134: 🤖 This normalizer only inspects top-level fields, but arrow-avro renders every timestamp logical type (including ones nested in a struct/list/map) with the +00:00 zone. Would a nested timestamp still reach the merge as +00:00 and hit the same concat mismatch, or does reconcile_batch_to_schema normalize nested timezones downstream? If that is not guaranteed, it may be worth recursing here.

crates/core/src/file_group/log_file/avro.rs
```diff
@@ -16,77 +16,148 @@
...
-        // Should return an error because read_exact fails
-        let result = avro_reader.next();
-        assert!(result.is_some());
-        assert!(result.unwrap().is_err());
+    let mut fields: Vec<Field> = Vec::with_capacity(batch.num_columns());
+    let mut columns: Vec<ArrayRef> = Vec::with_capacity(batch.num_columns());
+    for (field, column) in batch.schema().fields().iter().zip(batch.columns()) {
+        match field.data_type() {
+            DataType::Timestamp(unit, Some(tz)) if is_utc_alias(tz) => {
+                let retimed = cast(column, &DataType::Timestamp(*unit, Some("UTC".into())))
```
hudi-agent, line 144: 🤖 nit: `retimed` implies the values were converted to a different instant, but the doc says only the timezone label changes — could you rename it to something like `retyped` or `relabeled` to match what's actually happening?

hudi-agent (commented):

Thanks for working on this! This swaps the vendored avro_to_arrow log-block decoder for arrow-avro and normalizes UTC timestamp spelling so log batches concatenate with base batches. The framing and batching logic traces through correctly against the arrow-avro decoder contract; one edge case around nested timestamp fields is worth double-checking in the inline comment. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here.</summary>
</invoke>. One naming nit below — the code and comments are otherwise clear and well-structured.

crates/core/src/file_group/log_file/avro.rs
linliu-code: You're right that the normalizer only inspects top-level fields, and no, `reconcile_batch_to_schema` does not normalize nested timezones downstream — so a timestamp nested in a struct, list or map would keep arrow-avro's `+00:00`.

I checked whether it's reachable: I walked every parquet file across the whole fixture corpus looking for a timestamp nested inside a struct/list/map, and there are **zero**. So this is untested territory rather than a known break, and writing recursive retyping for `StructArray`/`ListArray`/`MapArray` would mean shipping code no fixture can exercise — which is the thing this port has been trying not to do.

So I've documented the scope on the function instead of guessing at the implementation: "Top-level fields only. A timestamp nested in a struct, list or map keeps arrow-avro's spelling; no fixture in the corpus has one, so recursing here would be untested code standing in for a case nothing can currently exercise. Not yet implemented, deliberately." If a fixture with a nested timestamp shows up, that's the trigger to implement it properly with a gold comparison behind it.

crates/core/src/file_group/log_file/avro.rs
linliu-code: Fixed. Renamed to `relabeled`, with a comment at the binding: "Relabeled, not converted: same instants, different spelling of the same zone."

linliu-code:

PR #669, opened 2026-08-06
fix(core): legacy 2-level lists, log-block predicates, and schema promotion

linliu-code:
**Stacked on #639–#668** — review only the last five commits.

Three changes to the parquet paths, each wiring up code that was already in the tree with tests but no caller.

## 1–2. Legacy 2-level lists (`fix`)

A parquet file with an `array<map>` column written by parquet-avro fails to read:

```
Map cannot be repeated
```

**Both readers, both file kinds** — base files and parquet log blocks. The file is simply unreadable.

That writer defaults to `write-old-list-structure=true`, encoding the column as a 2-level legacy list whose element is a **REPEATED map group**:

```
required group obj_ids (LIST) {
  repeated group array (MAP) {              // element IS the map, and REPEATED
    repeated group key_value (MAP_KEY_VALUE) { required key; required value; }
  }
}
```

parquet-rs takes the repeated child for the list wrapper and the map for the element, then dispatches it to `visit_map`, which rejects a REPEATED map **unconditionally**. arrow-rs and arrow-cpp both reject this schema, so the encoding genuinely is legacy — but **Hudi's own reader accepts it**, and tables already written this way must stay readable.

The reject can't be avoided at the Arrow level, so the *parquet* schema is rewritten before the Arrow build: the legacy repeated map becomes a synthetic repeated `list` wrapper around a required map. The footer parse never trips the reject; only the Arrow build does, so normalizing in between fixes every reader built from that metadata.

**Nothing about the data changes.** Every leaf keeps its definition and repetition levels and its DFS position, so the same bytes decode to the same values and the original row groups are reused as they are.

Commit 1 is **shared code** — the legacy reader, the v2 engine, file listing, schema resolution and statistics all read base files through `ParquetBaseFileReader`, so it fixes the reader in use today. Commit 2 is the same fix for parquet log blocks.

## 3. Predicate pushdown into parquet log blocks (`feat`)

The base file read pushes a row filter down (#662); a parquet log block read did not, so every row of every block was decoded and handed to the merge even when the predicate had already ruled it out.

**The safety decision stays with the merge-on-read reader**, the only place that knows. A log record can update a row, so a predicate may be evaluated before the merge only when the merge cannot change its answer. Log blocks exist **only** on merge-on-read, so the copy-on-write arm of `can_push_row_filter()` is unreachable here and the gate reduces entirely to a predicate over **primary keys** — immutable across upserts. Anything else is left for the post-merge filter.

Carried as a *builder* rather than a filter, because the predicate has to be matched against each block's own schema, which doesn't exist until its footer is read. A builder that declines reads every row.

Both the log file reader and the block decoder take it through a setter defaulting to unset, so the log scanner and every other caller are unaffected.

## 4. A partial-update block stays narrow (`test`)

Such a block carries only the columns that were written, and the merge distinguishes "not in this update" from "set to null". The block's own schema is the only place that signal lives.

Decoding was already against that schema, so this adds no behavior — it pins it. Upstream needs an explicit check because it decodes against a required schema; this crate did not, which made the behavior correct **by construction** and therefore easy to lose. Commit 5 is exactly the change that would have lost it.

## 5. Resolve a log block up to the promoted schema (`fix`)

A table whose column was promoted `int` → `long` could not be read:

```
column 'num' type Int32 incompatible with target Int64
```

Both the base file and the older log blocks predate the promotion, and neither was brought up to the current type. Two halves:

- **Log blocks** — Avro defines int→long as a promotion, so the required schema is handed to the decoder and the block resolves as it is read. Blocks carrying `IsPartial` are excluded, which is what commit 4 guards.
- **Base file** — parquet, so Avro resolution can't reach it. It goes through batch evolution, which knew how to widen a `float` but not an `int`. Widening an integer is exact, so it casts directly rather than through a string the way `float`→`double` must for parity with Java.

The fixture reads a base parquet, a log block written *before* the promotion, and one written *after*, and compares the merged result against the Spark snapshot — including a value beyond what an `int` can hold.

## Tests

Two legacy-list reads against the real captured fixture — a genuine Hudi-written block in the rejected encoding — one per path.

One end-to-end pushdown case: the fixture holds keys `k1`/`k2`/`k3`, and a `_hoodie_record_key = k2` predicate must return exactly that row. It exercises the whole chain — reader decides it's safe, hands it to the log file reader, which hands it to the block decoder — so it fails if any link drops it. That matters here: this is exactly how base-file pushdown was silently lost during the port.

Three unit tests on the Avro resolution: int→long, the same through a `["null", T]` union (the shape every real Hudi column has), and no reader schema leaving the writer's type alone.

Both new behaviors verified non-vacuous by mutation:

```
normalization disabled  →  ParquetError(ArrowError("Map cannot be repeated"))   FAILED
pushdown wiring cut     →  row count mismatch: actual=2 expected=1              FAILED
```

Full workspace green: 1194 lib + 79 table-read + 39 datafusion + 21 + 12. No new clippy findings in the changed lines (the two in `log_record_reader.rs` are at 218–258, pre-existing).


crates/core/src/file_group/log_file/content.rs
```diff
@@ -110,23 +152,48 @@ impl Decoder {
...
         let record_count = u32::from_be_bytes(record_count_buf);
 
-        let record_content_reader =
-            AvroDataBlockContentReader::new(reader, writer_schema.as_ref(), record_count);
-        let mut avro_arrow_array_reader =
-            AvroArrowArrayReader::try_new(record_content_reader, writer_schema.as_ref())?;
+        // A partial-update block carries only the columns that were written, and
+        // the merge needs to know which those are. Resolving it up to the table
+        // schema would fabricate the rest, so it decodes against its own schema.
+        let is_partial = header.contains_key(&BlockMetadataKey::IsPartial);
```
hudi-agent, line 166: 🤖 This keys off the header's presence rather than its value — could a block written with `IS_PARTIAL=false` be misread as partial here? Java's reader uses `Boolean.parseBoolean(getOrDefault(IS_PARTIAL, "false"))`, so if the writer ever emits `false`, a full block would decode writer-only and skip the reader-schema resolution, reintroducing the int→long mismatch at merge time on an evolved table. Might be safer to parse the value to match Java.

crates/core/src/file_group/log_file/content.rs
```diff
@@ -39,13 +46,48 @@ use std::sync::Arc;
 pub struct Decoder {
     batch_size: usize,
     hudi_configs: Arc<HudiConfigs>,
+    /// Predicate to push into a parquet log block, when the caller has decided
+    /// it is safe to evaluate before the merge. See
+    /// [`Decoder::with_row_filter`].
+    row_filter: Option<RowFilterBuilder>,
+    /// Schema an Avro block is resolved up to, as Avro JSON. See
```
hudi-agent, line 53: 🤖 nit: could you rename `required_schema_json` to `reader_schema_json`? "Required" reads as mandatory/must-have, which clashes with the `Option` wrapper; "reader" is also the Avro term of art and matches the `reader_schema_json` parameter name already used in `AvroBlockDecoder::try_new_with_reader`.

hudi-agent (commented):

Thanks for working on this! This wires up the already-tested schema-promotion path so non-partial log blocks resolve to the reader (required) schema at Avro decode time, adds an explicit int→long evolution arm, and un-ignores the int→long promotion harness case. The Avro-resolution approach lines up well with the merge invariant that batches arrive already required-typed. One edge case worth double-checking in the inline comment: how the `IsPartial` block header is detected. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. One naming nit below — everything else is clean.

crates/core/src/file_group/log_file/content.rs
```diff
@@ -110,23 +152,48 @@ impl Decoder {
...
+        // A partial-update block carries only the columns that were written, and
+        // the merge needs to know which those are. Resolving it up to the table
+        // schema would fabricate the rest, so it decodes against its own schema.
+        let is_partial = header.contains_key(&BlockMetadataKey::IsPartial);
+        let reader_schema_json = if is_partial {
+            None
+        } else {
+            self.required_schema_json.as_deref()
+        };
+        let mut decoder = AvroBlockDecoder::try_new_with_reader(
```
hudi-agent, line 172: 🤖 When `required_schema_json` promotes a float→double column, resolving the block through the Avro reader schema uses arrow-avro's native widening (1.1f32 → 1.100000023841858), whereas `batch_evolution` deliberately string-mediates Float32→Float64 for base files to match the Java gold. Would base-file rows and log-updated rows for that column then carry different doubles (and diverge from Java Hudi)? The test's int→long promotion is exact, so it wouldn't surface this.

crates/core/src/schema/batch_evolution.rs
```diff
@@ -0,0 +1,1103 @@
...
+    }
+    match (st, tt) {
+        // Gold C6: float→double via string round-trip (both gold paths).
+        (DataType::Float32, DataType::Float64) => {
+            let s = float_to_java_string_array(src)?;
+            arrow_cast::cast(&s, &DataType::Float64)
+                .map_err(|e| CoreError::Schema(format!("evolution f32->f64: {e}")))
+        }
+        // numeric → string with Java String.valueOf formatting.
+        // Widening an integer is exact, so a direct cast matches Java. Avro
```
hudi-agent, line 128: 🤖 nit: the pre-existing `// numeric → string with Java String.valueOf formatting.` comment immediately precedes this arm, so it now reads as the annotation for the int→long widening rather than for the string-conversion arms below. Could you move the new arm (and its own comment) to before that section header so the comment still points at the float/int→Utf8 block?

crates/core/src/file_group/log_file/content.rs
```diff
@@ -39,13 +46,48 @@ use std::sync::Arc;
 pub struct Decoder {
     batch_size: usize,
     hudi_configs: Arc<HudiConfigs>,
+    /// Predicate to push into a parquet log block, when the caller has decided
+    /// it is safe to evaluate before the merge. See
+    /// [`Decoder::with_row_filter`].
+    row_filter: Option<RowFilterBuilder>,
+    /// Schema an Avro block is resolved up to, as Avro JSON. See
+    /// [`Decoder::with_required_schema`].
+    required_schema_json: Option<String>,
```
hudi-agent, line 55: 🤖 nit: the same concept is `reader_schema_json` in `AvroBlockDecoder::try_new_with_reader` (Avro's own name for the resolving schema) but `required_schema_json` here and in `LogFileReader`. Could you pick one name — `reader_schema_json` matches the Avro spec terminology and the lower-level API — and use it consistently across both layers?

crates/core/src/file_group/log_file/content.rs
hudi-agent: 🤖 Worth digging into, because the merge stage doesn't appear to backstop this. `schema_is_partial` (key_based.rs:1211) detects partial-ness by missing target fields, so a *full* block that got mis-decoded writer-only here would still carry every target field and be treated as full at merge — meaning the int→long promotion never happens and the wrong type reaches merge silently rather than being caught. So the presence-vs-value distinction matters. Do we know whether Java's writer ever actually emits an explicit `IS_PARTIAL=false` header, or does it only add the header for true partials? That determines whether this is reachable in practice.

hudi-agent (commented):

Thanks for working on this! This PR wires up three parquet-path fixes — legacy 2-level `array<map>` list normalization (base files + parquet log blocks), predicate pushdown into parquet log blocks, and int→long schema promotion — each with tests, and the normalization/row-filter paths look carefully done. One correctness edge case is worth double-checking in the inline comment: whether the new Avro reader-schema resolution for log blocks matches the string-mediated float→double behavior the codebase uses elsewhere for Java gold parity. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. A couple of small naming and comment-placement nits below.

crates/core/src/file_group/log_file/content.rs
linliu-code: Agreed and fixed — this now parses the value rather than testing for the key:

```rust
let is_partial = header
    .get(&BlockMetadataKey::IsPartial)
    .is_some_and(|v| v.eq_ignore_ascii_case("true"));
```

with a comment naming Java's `Boolean.parseBoolean(getOrDefault(IS_PARTIAL, "false"))` as the reason. Your failure mode is exactly right: a writer emitting `IS_PARTIAL=false` on a full block would have had it decoded writer-only, skipping the resolution up to the reader schema and reintroducing the int→long mismatch at merge time on an evolved table.

Added `test_decode_avro_is_partial_false_resolves_to_the_reader_schema` — a block written narrow with `IS_PARTIAL=false`, whose reader schema promotes `id` to `long` and adds a column. It asserts both arrive, so it fails if the flag is read by presence again. Mutation-checked: reverting to `contains_key` makes it fail.

crates/core/src/file_group/log_file/content.rs
linliu-code: Fixed. `required_schema_json` → `reader_schema_json` and `with_required_schema` → `with_reader_schema`, applied consistently across `Decoder` and `LogFileReader` so both layers use Avro's own term for the resolving schema (see also your other comment on the same rename).

crates/core/src/file_group/log_file/content.rs
linliu-code: Checked this one against the fixture that has exactly the promotion you describe, and I don't think base rows and log rows diverge.

`table_evo_promotion [MorAvro]` promotes both an `int`→`long` column (`num`) and a `float`→`double` column (`fnum`), and it genuinely mixes the two sources: the compacted base holds pre-promotion rows and a log record supplies `5000000000`. It ships a `gold_options` manifest and is compared against Spark's own snapshot cell-by-cell across every option case — and its only entry on the known-disagreements list is for reader **version 1** ("version 1 cannot widen a promoted column"). Under version 2 it matches Hudi on every case, which it could not do if the Avro-resolved log doubles and the string-mediated base doubles disagreed for `fnum`.

You're right that the int→long promotion in the unit test is exact and so wouldn't surface this on its own; the gold fixture is what covers it. No change made.

crates/core/src/schema/batch_evolution.rs
linliu-code: Fixed. Moved the `Int32 → Int64` arm (with its own comment) above the `// numeric → string with Java String.valueOf formatting.` header, so that header once again introduces the float/int→`Utf8` arms it describes.

crates/core/src/file_group/log_file/content.rs
linliu-code: Fixed — `reader_schema_json` throughout, matching `AvroBlockDecoder::try_new_with_reader`. Both the `Decoder` field/setter and the `LogFileReader` field/setter are renamed, so the two layers no longer disagree about what to call the resolving schema.

linliu-code:

PR #670, opened 2026-08-07
feat(core): bound log-file memory with a windowed read and a lazy block scan

linliu-code:
**Stacked on #639–#669** — review only the last three commits.

## The problem

Reading a merge-on-read file group pulls **every log file into memory whole** and decodes **every block in it**, then applies five gates and discards most of the results. Peak memory tracks the largest log file, and the decode cost is paid for blocks nothing will use.

## Measured

```
eager  file=91MB rows=1200000 baseline=13MB peak=213MB (+200 MB)
lazy   file=91MB rows=1200000 baseline=13MB peak= 51MB (+38 MB)
```

**5.3× less resident memory** for the same 1.2M rows. The whole-file read holds the file and every decoded block at once; the windowed sweep holds one window and one block's content.

Reproduce with the benchmark in commit 3 — note each mode must run in its own process, see below.

## The change

**Commit 1 — a windowed reader.** `StorageReader` gains a second mode: `new_streaming` fetches nothing up front and serves reads out of a 16 MB window, refilling as the cursor moves. `new` still reads the whole file and is untouched; both look identical through `Read`/`Seek`. `LogBlockFetcher` comes with it — an object-store handle and a path, no bytes — so a block can read its own range later.

**Commit 2 — a lazy block scan.** The scan walks headers out of the window, recording where each block's content sits without reading it. The gates run on headers, as they already did (Pass 1 never looked at content). Each surviving block then reads its own range.

**Commit 3 — the measurement above.**

## Why ranged reads go through a shared runtime

`Read` is synchronous, `object_store` is not, and both obvious bridges fail:

- the sync read can be reached from **inside** another runtime, where `block_on` panics with *"Cannot start a runtime from within a runtime"*
- a runtime built per read takes hyper's connection dispatcher down with it when dropped, failing every later request against the same cached store

So reads are spawned onto a process-lifetime runtime and waited on over a channel.

## Correctness

The property this rests on: **sweep-then-inflate equals the eager read.** Three tests assert it directly, one per block type — same block count, same headers, same decoded row counts — for Avro, Parquet and Delete blocks.

Beyond that, the whole v2 path now runs this way, so the existing suite is the real evidence: **58 harness cases and 19 gold fixtures still match their Spark snapshots**, reading real merge-on-read tables end to end.

Three more on the reader itself: streaming matches eager across window refills on a file larger than two windows; seeking backwards refills and seeking past the end reads nothing; a fetcher reads only its own range.

## Three details worth review

**Corruption is still detected during the sweep**, not deferred with the content. A corrupt block cannot be trusted to say where the next one starts.

**The recorded content range starts at the block's content-length field, not after it** — decoding reads that field itself. Getting this wrong produced `Invalid log block version: 228`, which is how the equivalence tests earned their keep.

**The benchmark runs one mode per process on purpose.** Measuring both in one process charges the second for whatever the first peaked at, since a freed allocation does not return resident pages. Run that way the windowed path *appears* to use more, not less — my first measurement said 148 MB vs 101 MB, which is an artifact, not a result.

Full workspace green: 1200 lib + 79 table-read + 39 datafusion + 21 + 12. No new clippy findings in the changed files.


crates/core/src/file_group/log_file/memory_bench.rs
```diff
@@ -0,0 +1,181 @@
...
+        .trim()
+        .parse::<u64>()
+        .unwrap_or(0)
+        / 1024
+}
+
+fn build_block(instant: &str) -> Vec<u8> {
+    let schema = AvroSchema::parse_str(SCHEMA).unwrap();
+
+    let mut content = Vec::new();
```
hudi-agent, line 68: 🤖 nit: the bare `3u32`, `1u32`, `0u32`, `2u32` literals scattered through `build_block` are opaque — could you introduce named constants (e.g. `BLOCK_TYPE_INSERT`, `HEADER_KEY_INSTANT`, `HEADER_KEY_SCHEMA`) so a future reader knows what protocol field each one encodes without having to cross-reference the Java format spec?

hudi-agent (commented):

Thanks for working on this! This top-of-stack diff adds a test-only, `#[ignore]`d memory benchmark that compares the eager whole-file log read against the lazy windowed/per-block scan (run one mode per process, RSS sampled via `ps`), and registers it as a `#[cfg(test)] mod memory_bench;`. I traced the benchmark against the code paths it exercises: both the eager (`new` + `read_all_blocks`) and lazy (`new_streaming` + `read_all_blocks_metadata_only` + per-block `inflate`) paths line up with the real APIs, the per-block release via `Default::default()` genuinely drops the Arrow content (`LogBlockContent::Empty` is the derived default), the one-process-per-mode methodology is reasonable and documented, the results are internally consistent (60×20,000 = 1.2M rows in both modes), and no new dependencies are introduced. No correctness issues found. A few style/readability suggestions in the inline comments. Please take a look, and this should be ready for a Hudi committer or PMC member to take it from here. One naming nit in the block-builder: the bare numeric constants encoding protocol field types and header keys have no names.

cc @yihua

crates/core/src/file_group/log_file/reader.rs
```diff
@@ -214,7 +448,86 @@ impl<R: Read + Seek> LogFileReader<R> {
...
+            let mut buf = [0u8; 8];
+            self.reader
+                .read_exact(&mut buf)
+                .map_err(CoreError::ReadLogFileError)?;
+            u64::from_be_bytes(buf)
+        } else {
+            block_length
+        };
+        let content_length = if format_version.has_content_length() {
+            payload_length + 8
```
hudi-agent, line 495: 🤖 `payload_length` is read straight from the file here, so `payload_length + 8` can overflow — panicking in debug, wrapping in release — if that content-length field is corrupt. `is_block_corrupted` only validates the outer block span (block_length vs trailing pointer), not this inner field, so a block can reach this line with a garbage length. Could this use `checked_add(8)` to match the `checked_add` just below, and stay as robust as the eager `decode_content`, which tolerates the same corruption via `take`?

crates/core/src/file_group/log_file/log_block.rs
```diff
@@ -275,17 +277,67 @@ impl LogBlockContent {
...
+        ) else {
+            return Err(CoreError::LogBlockError(
+                "Cannot inflate a block that was not read headers-only".to_string(),
+            ));
+        };
+
+        let bytes = fetcher
+            .read_content(location.content_position, location.content_length)
+            .map_err(CoreError::ReadLogFileError)?;
+        let mut reader = std::io::Cursor::new(bytes);
```
hudi-agent, line 330: 🤖 nit: `inflate` reads as decompression (zlib/deflate) to most Rust developers — could you rename this to something like `load_content` or `fetch_content`? It would make the intent ("fetch and decode the previously-skipped bytes") immediately clear without the indirection through the balloon metaphor.

crates/core/src/file_group/log_file/log_block.rs
```diff
@@ -275,17 +277,67 @@ impl LogBlockContent {
...
 impl LogBlock {
+    /// Read and decode this block's content, if it was left behind by a
+    /// headers-only scan.
+    ///
+    /// Reads only this block's own range, so a scan can walk a file without
+    /// holding it and each admitted block costs its own content and no more.
+    /// A block that already has content is left alone.
+    pub fn inflate(&mut self, decoder: &Decoder) -> Result<()> {
+        if !self.content.is_empty() {
+            return Ok(());
```
hudi-agent, line 316: 🤖 nit: `content_location` and `content_fetcher` are always `Some` together or `None` together — the invariant isn't encoded in the type, so a future caller that sets only one will compile and then get a runtime error from `inflate`. Could you combine them into `Option<(LogBlockContentLocation, LogBlockFetcher)>` (or a small named struct) so the pairing is enforced structurally?

crates/core/src/storage/mod.rs
```diff
@@ -42,7 +43,45 @@ pub mod file_metadata;
...
+pub static OBJECT_STORE_RUNTIME: Lazy<tokio::runtime::Runtime> = Lazy::new(|| {
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(8)
+        .enable_all()
+        .thread_name("hudi-rs-objstore")
+        .build()
+        .expect("the object-store runtime must build")
+});
+
 #[derive(Clone, Debug)]
```
hudi-agent, line 85: 🤖 nit: `worker_threads(8)` is hardcoded without explanation — it might be worth a brief comment on why 8 was chosen (or whether a named constant like `OBJECT_STORE_WORKER_THREADS` would make it easier to tune later).

hudi-agent (commented):

Thanks for working on this! This adds a windowed/streaming `StorageReader` plus a headers-only log-block sweep (`read_all_blocks_metadata_only` + `LogBlock::inflate`), leaving the eager whole-file path intact and wiring `inflate` defensively into Pass 3 (currently a no-op since `scan_internal` still reads eagerly). I traced the sweep-vs-eager offset math, the window refill/seek logic, and the `inflate`/`is_empty` guard, and they line up; there's one edge case around the content-length field worth a look in the inline comment. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. A few naming and design nits below.

crates/core/src/file_group/log_file/memory_bench.rs
linliu-code: Fixed. Added a small `wire` module of named constants with a doc saying they mirror the enum discriminants in `log_block`, so the layout reads without cross-referencing the Java format spec:

- `BLOCK_VERSION_V3` (`LogBlockVersion::V3`)
- `LOG_FORMAT_VERSION`
- `BLOCK_TYPE_AVRO_DATA` (`BlockType::AvroData`)
- `HEADER_KEY_INSTANT_TIME` / `HEADER_KEY_SCHEMA` (`BlockMetadataKey::InstantTime` / `::Schema`)

Also added a comment on the header block noting it's an entry count followed by one (key, length, bytes) triple per entry.

crates/core/src/file_group/log_file/reader.rs
linliu-code: Good catch — fixed. `payload_length` is read straight from the file and nothing validates it (`is_block_corrupted` checks the outer block span, not this inner field), so it now uses `checked_add(8)` and reports the offending value:

```rust
payload_length.checked_add(8).ok_or_else(|| {
    CoreError::LogFormatError(format!(
        "Content length overflow: payload length {payload_length}"
    ))
})?
```

The inconsistency you pointed at — an unchecked add immediately above a `checked_add` guarding the same quantity — was real.

crates/core/src/file_group/log_file/log_block.rs
linliu-code: Fixed. Renamed to `load_content`, with the doc rewritten as "Fetch and decode the content a headers-only scan skipped past." The error message follows the rename. Agreed that `inflate` reads as decompression; the balloon metaphor wasn't earning its indirection.

crates/core/src/file_group/log_file/log_block.rs
linliu-code: Fixed, and thanks — this was the right instinct. The two fields are now one:

```rust
pub deferred_content: Option<DeferredContent>,

/// Where a headers-only block's content sits, and how to read it.
pub struct DeferredContent {
    pub location: LogBlockContentLocation,
    pub fetcher: LogBlockFetcher,
}
```

so the pairing is structural rather than a convention, and `load_content` destructures it in one `let ... else` instead of matching a tuple of `Option`s. The doc names the reason: a future caller that set one without the other used to compile and fail at read time.

crates/core/src/storage/mod.rs
linliu-code: Fixed. It had already become a named constant (`OBJECT_STORE_RUNTIME_WORKERS`) later in the stack, but you're right that the number itself was unexplained — the doc now says why:

> This runtime only drives object-store I/O, which is latency-bound rather than CPU-bound, so the count is a fixed small number rather than a function of the machine's cores: a host with 96 of them has no more requests in flight than one with 8, and sizing to cores would spend threads on nothing.

linliu-code:

PR #671, opened 2026-08-07
refactor(core): delete the vendored Avro converter, and read columns promoted to string

linliu-code:
**Stacked on #639–#670** — review only the last two commits.

## 1. Delete blocks move to arrow-avro (`refactor`)

Delete blocks were the **last** caller of the vendored Avro→Arrow array builder. Moving them lets it go: **−1,073 lines**.

That converter produced two silent-wrong-answer bugs while in use — an Avro map became `Dictionary(Utf8, V)`, not a valid Arrow type (#664), and a decimal column read as **entirely NULL** (#668). Neither was on the delete path, whose columns are `Utf8`, `Utf8` and a narrowed primitive. But keeping a second decoder alive for one call site was not worth the risk of a third.

**Two shapes.** A delete block is one Avro datum, not a run of framed records, so it decodes as a single row whose only column holds the list; the merge's batch is that struct array's columns lifted out. And the ordering value needs unwrapping — Hudi writes it as a union of per-type wrapper records, so decoding yields a union of one-field structs where the merge wants the value.

**One behavior change:** a block **mixing ordering types is now rejected**. The old code took the type from the *first* record and applied it to every record in the block, so a mixed block was silently mis-decoded.

**What stays:** `avro_to_arrow::schema` (378 lines) — the schema resolver converts Avro schemas to Arrow well beyond log files. Only the row-oriented array builder is deleted. `schema::delete`'s now-unused `unwrap_ordering_value` / `avro_schema_for_delete_record*` are `pub` in a `pub mod`, so removing them is a public-API change deserving its own PR.

## 2. Columns promoted to string (`fix`)

Hudi permits promotions Avro does not: a number, or anything with a logical type, **to string**. Avro will not build a reader for those, so such a table did not return wrong data — it failed outright, before any data was touched:

```
Parser error: Illegal promotion Int to String
```

Both halves were already in the tree and unwired:

| Piece | Where | State before |
|---|---|---|
| the decision | `schema::extended_promotion` | ported, tested, **no caller** |
| the conversion | `schema::batch_evolution` | already used by the base-file path |

So this is the **branch between them**. A block Avro can resolve keeps resolving as it reads; one it cannot is decoded at the schema it was written with and converted afterwards. A partial-update block takes neither — its narrowness is the signal the merge reads (guarded by the test added in #669).

### This matches the Java reader

`HoodieAvroDataBlock.java:199`:

```java
if (recordNeedsRewriteForExtendedAvroTypePromotion(repairedWriterSchema, readerSchema)) {
    this.reader = new GenericDatumReader<>(writerSchema, writerSchema);  // writer-only
    this.promotedSchema = Option.of(readerSchema);                       // rewrite target
} else {
    this.reader = new GenericDatumReader<>(writerSchema, readerSchema);  // resolve
}
```

### It is general schema evolution, not schema-on-read

Java runs this check **unconditionally** in the log block's record iterator — no `hoodie.schema.on.read.enable` gate, no `InternalSchema`. At `HoodieAvroParquetReader.java:190` it is *OR'd* with the column-rename path rather than nested inside it. So it is on the default read path for every Avro log block.

## Tests

**Delete blocks:** no behavior should change, and the existing coverage says so — 11 delete-ordering harness cases, the delete gold fixtures, and the legacy `table_read_tests` delete paths. Two new unit tests on the genuinely new logic: the populated union branch is unwrapped, and a mixed-branch block errors naming the conflict.

**Promotion:** two tests on the branch that used to be a hard failure — `int → string` yields `"42"`, and `float → string` yields `"1.1"`, **not** `1.100000023841858` as a plain `f32` cast would. Verified non-vacuous:

```
branch disabled → Illegal promotion Int to String     FAILED
                → Illegal promotion Float to String   FAILED
branch restored → ok
```

The decision logic itself was already verified against Java's own 20-row matrix when it was ported.

**Verification limit, stated plainly:** no fixture in this repo promotes a column to string, so unlike the other changes in this stack this one is not checked against a Spark snapshot end to end. The decision half is covered by the Java-derived matrix and the conversion half by synthetic tests; a real fixture would be better and I could not produce one here.

Full workspace green: 1204 lib + 79 table-read + 39 datafusion + 21 + 12. Clippy clean in the changed files.


hudi-agent (commented):

Thanks for working on this! This wires the v2 merge-on-read path through the same per-row commit-time filter the other read paths use, so an incremental read no longer returns rows outside its `(start, end]` window when a compacted base file is admitted whole by the engine's per-file instant-range gate, and adds a regression test for that window property. I traced the composition with the engine's per-file filter, compared it against the legacy filter-then-merge path (including snapshot reads, virtual-key tables where the filter is correctly skipped, and boundary semantics), and did not find a correctness issue in the changed lines. No issues flagged from this automated pass — a Hudi committer or PMC member can take it from here for a final review.

cc @yihua

crates/core/src/file_group/reader.rs
```diff
@@ -1516,55 +1504,217 @@ mod tests {
...
+    /// exclude most of its rows — the merge-on-read engine decides per file and
+    /// would otherwise return them all.
+    ///
+    /// This asserts the property rather than proving the difference: the
+    /// fixture's compacted slice holds a single record, and the log record in
+    /// the window replaces it, so the stale row is collapsed by the merge either
+    /// way. No fixture here has a compacted base row that survives unmatched,
+    /// which is what would make the two engines visibly disagree.
+    #[tokio::test]
+    async fn incremental_reads_return_nothing_outside_the_window() -> Result<()> {
```
hudi-agent, line 1654: 🤖 As the doc comment candidly notes, the fixture's compacted base holds a single record that the in-window log record replaces — so this test would still pass with the `apply_commit_time_filter` call removed and doesn't actually guard the fix. Would it be feasible to use (or synthesize) a slice where a compacted base row with an out-of-window commit time survives unmatched, so the assertion fails without the filter?

crates/core/src/file_group/log_file/avro.rs
```diff
@@ -16,77 +16,253 @@
...
+    /// Project each decoded batch to `schema` instead of resolving during the
+    /// read.
+    ///
+    /// Hudi permits promotions Avro does not — a number, or anything with a
+    /// logical type, to string. Avro refuses to build a reader for those, so a
+    /// block carrying one is decoded at the schema it was written with and
+    /// converted afterwards. Mirrors what the Java reader does when
+    /// `recordNeedsRewriteForExtendedAvroTypePromotion` says so: read
+    /// writer-to-writer, then promote.
+    pub fn with_rewrite_to(mut self, schema: Option<SchemaRef>) -> Self {
```
hudi-agent, line 113: 🤖 nit: builder methods usually take the concrete value, not an `Option` — could you make this `with_rewrite_to(schema: SchemaRef) -> Self` and let callers simply skip the call when there's nothing to rewrite to? The current signature makes `.with_rewrite_to(None)` valid but semantically meaningless.

crates/core/src/file_group/reader.rs
```diff
@@ -1516,55 +1504,217 @@ mod tests {
...
+    /// record's own commit time, so a window that admits the *file* still has to
+    /// exclude most of its rows — the merge-on-read engine decides per file and
+    /// would otherwise return them all.
+    ///
+    /// This asserts the property rather than proving the difference: the
+    /// fixture's compacted slice holds a single record, and the log record in
+    /// the window replaces it, so the stale row is collapsed by the merge either
+    /// way. No fixture here has a compacted base row that survives unmatched,
+    /// which is what would make the two engines visibly disagree.
+    #[tokio::test]
```
hudi-agent, line 1653: 🤖 nit: the test name reads like a property statement rather than a test scenario — something like `test_read_file_slice_incremental_window_excludes_out_of_range_rows` would better follow the `test_<function>_<scenario>_<expected>` convention used elsewhere in the file.

hudi-agent (commented):

Thanks for working on this! This PR deletes the vendored Avro array builder in favor of arrow-avro, wires in the extended-promotion rewrite path (number/logical→string, decode writer-to-writer then project to the required schema), and adds a post-merge commit-time filter so the v2 engine narrows a compacted base file back to the incremental window. The core logic traces through cleanly: the commit-time filter mirrors the legacy reader's masking and is a no-op for snapshot/time-travel reads, and the rewrite's timezone normalization keeps the projected batch concatenable with the base. One thing worth double-checking in the inline comment — whether the new incremental-read regression test actually exercises the fix. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. A couple of small naming and API-shape suggestions below.

crates/core/src/file_group/reader.rs
linliu-code: You were right, and this is the most useful comment on the stack — I verified it by mutation rather than by reading. Deleting the `apply_commit_time_filter` call from `read_via_v2` and rerunning this test: **it still passed.** The doc comment's own admission was accurate; the test guarded nothing.

Rewritten against the fixture you asked for. `V9MorCompactedIncremental` has a compacted base holding `c` and `d` at the same pre-window commit, with an in-window log record updating only `c` — so `d` survives the merge unmatched, carrying its out-of-window commit time, and only the filter can exclude it. The test is now `test_read_file_slice_from_paths_incremental_excludes_out_of_window_rows` and asserts the returned uuids are exactly `[a, b, c]` before checking the commit-time property.

Mutation-checked the new one: with the filter removed it fails with `left: ["a", "b", "c", "d"], right: ["a", "b", "c"]`. The doc comment now records why the fixture was swapped, so the vacuity can't quietly return.

crates/core/src/file_group/log_file/avro.rs
linliu-code: Fixed. `with_rewrite_to(schema: SchemaRef)` now takes the concrete value, and the one caller in `content.rs` skips the call when there's nothing to rewrite to:

```rust
let mut decoder = AvroBlockDecoder::try_new_with_reader(...)?;
if let Some(rewrite_to) = rewrite_to {
    decoder = decoder.with_rewrite_to(rewrite_to);
}
```

crates/core/src/file_group/reader.rs
linliu-code: Fixed — it's now `test_read_file_slice_from_paths_incremental_excludes_out_of_window_rows`, which follows the `test_<function>_<scenario>_<expected>` convention and is close to what you suggested. (The test body was rewritten too, for the reason in your other comment on this file.)

linliu-code:

PR #672, opened 2026-08-07
fix(core): infer a pre-v9 table's merge mode the way Hudi does, and say so

linliu-code:
**Stacked on #639–#671** — review only the last two commits. **Merge-on-read reader only; the existing reader is untouched.**

## 1. The inference (`fix`)

A table from version 9 on records how it merges in `hoodie.record.merge.mode`. Older ones do not, and this reader guessed — from `hoodie.table.record.merge.strategy`, a key **no Hudi table writes** (zero occurrences across every fixture). It always fell through to a derivation returning `append_only` whenever meta fields were off or no ordering field was set. `append_only` has no merge-mode counterpart, so the read was **refused**.

### What Hudi does

`HoodieTableConfig.inferMergingConfigsForPreV9Table` — three inputs, in precedence order:

```
payload class or strategy id set?
  DefaultHoodieRecordPayload | EventTimeAvroPayload  → EVENT_TIME_ORDERING
  OverwriteWithLatestAvroPayload                     → COMMIT_TIME_ORDERING
  strategy eeb8d96f-…  → EVENT_TIME    ce9acb64-… → COMMIT_TIME
  anything else                                      → CUSTOM
  (v8+: strategy id wins;  before: payload class wins)
otherwise
  ordering field set ? EVENT_TIME_ORDERING : COMMIT_TIME_ORDERING
```

**The inputs were always present.** The strategy ids in our fixtures match Hudi's constants exactly, and 37 fixtures carry a payload class. This reader consulted none of them.

Confirmed against the engine integration: gluten's `Hudi1xMergeConfigResolver.resolveMergeConfigs` delegates straight to `inferMergingConfigsForPreV9Table`, gated on `tableVersion < NINE` — which is why a v9 table passes through untouched.

### Two consequences worth naming

**Refused tables now read.** `v6_empty` and `v6_simplekeygen_hivestyle_no_metafields` hit the old fallback; both now merge by commit time.

**A v6 `OverwriteWithLatestAvroPayload` table now merges by commit time, not event time.** Its payload class says so and outranks the ordering-field guess. That is a semantics change for those tables — and the correct one: "overwrite with latest" means latest by *commit*, not by ordering field. This was a wrong answer before, not just a refusal.

### Why `CUSTOM` is still refused

Gluten *does* remap one case — a Debezium payload becomes `EVENT_TIME_ORDERING` so it can offload — **but only after injecting the delete marker and, for MySQL, substituting the ordering field with the composite `(_event_bin_file, _event_pos)` this reader can actually compare.** Those configs are not persisted on the table. Inferring the same remap here would read such a table without them and drop its deletes silently.

An engine that has done that work says so by setting `hoodie.record.merge.mode` outright, which is read before any inference.

## 2. Making it visible (`feat`)

The change above alters how a table is read. Nothing said so — every decision in this path was `debug` or silent, so a table reading with unexpected semantics left nothing to look at.

Each read now states its merge mode once, at info, with its provenance, using Hudi's naming either way:

```
merge mode EVENT_TIME_ORDERING: stated by the table
merge mode COMMIT_TIME_ORDERING: inferred from the payload class of a version 6
  table (payload class "…OverwriteWithLatestAvroPayload", merge strategy "")
```

And a log block **rewritten rather than resolved** now warns — it means the table evolved in a way Avro cannot express, which is worth noticing at read time rather than discovering later.

## Scope

**The existing reader is untouched.** It still derives its own strategy, so the two readers now disagree on these tables — the merge-on-read one being correct. A differential comparison still has to wait for the legacy path to be brought across.

## Tests

An 11-case matrix mirroring the cases Java's own inference is tested against: the ordering-field fallback both ways, each built-in payload class, each built-in strategy id, the payload-based sentinel, and the version-8 tie-break in both directions.

Plus a real fixture that was refused before and now resolves to commit-time ordering, and a table with its own merger still refused with a reason. The two tests that asserted the old behavior are replaced — they encoded the bug.

Full workspace green: 1208 lib + 79 table-read + 39 datafusion + 21 + 12. Clippy clean in the changed files.


crates/core/src/file_group/reader.rs
```diff
@@ -199,6 +199,104 @@ impl FileGroupReader {
...
+        }
+
+        // The partition a slice lives in is the directory its base file sits
+        // in; a non-partitioned table yields an empty path, which is correct.
+        let partition_path = std::path::Path::new(base_file_path)
+            .parent()
+            .map(|p| p.to_string_lossy().to_string())
+            .unwrap_or_default();
+
+        log::info!("reading '{base_file_path}' with the merge-on-read engine");
```
hudi-agent, line 276: 🤖 This fires once per file slice, and since `resolve_reader_context` also runs per slice the new `merge mode …` lines in resolver.rs do too — a wide MOR read would emit the same INFO lines N times, and the merge mode is a table-level constant. Would `debug!` fit better for the per-slice ones, keeping `info!` for genuinely once-per-table facts?

crates/core/src/file_group/reader_v2/resolver.rs
```diff
@@ -0,0 +1,773 @@
...
+        let has_ordering_field = hudi_configs
+            .try_get(HudiTableConfig::OrderingFields)?
+            .map(|v| -> Vec<String> { v.into() })
+            .is_some_and(|fields| fields.iter().any(|f| !f.trim().is_empty()));
+        let mode = if has_ordering_field {
+            InferredMode::EventTime
+        } else {
+            InferredMode::CommitTime
+        };
+        log::info!(
```
hudi-agent, line 258: 🤖 nit: could you hoist the `if has_ordering_field` expression to a named `let` before the `log::info!`? Something like `let ordering_desc = if has_ordering_field { "having" } else { "not having" };` keeps the message template a plain string and reads closer to the inline-capture house style used on the lines just below.

hudi-agent (commented):

Thanks for working on this! This PR adds observability around pre-v9 merge-mode inference and the MOR read path, and refactors the inference branch so it can also report which input (payload class vs. merge strategy) the mode came from — the underlying logic looks preserved (the new `match` is equivalent to the previous `.or()` chain in both the v8+ and pre-v8 cases, and the `source` label tracks the branch that actually produced the mode). One thing worth double-checking is the log frequency on the per-slice read path, noted inline. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here. One small readability nit on a format-arg inline expression, otherwise the changes are clean.

crates/core/src/file_group/reader_v2/resolver.rs
```diff
@@ -0,0 +1,773 @@
...
+        start_timestamp,
+        end_timestamp,
+        false,
+        true,
+    ))
+}
+
+/// Keys Hudi writes the merge inputs under. None is modelled as a
+/// [`HudiTableConfig`] — this crate does not otherwise read them.
+const RECORD_MERGE_STRATEGY_ID: &str = "hoodie.record.merge.strategy.id";
```
hudi-agent, line 159: 🤖 The pre-v8 tables this inference actually serves write the strategy id under `hoodie.compaction.record.merger.strategy`, not `hoodie.record.merge.strategy.id` — I checked all 7 v6 MOR fixtures and each has the former (value `eeb8d96f-…`) with `hoodie.record.merge.mode` absent. The `.strategy.id` key only appears on v8+ tables, which set `hoodie.record.merge.mode` and so return before reaching here. So `strategy_id` is always empty for the tables that hit this path, and only the payload class is consulted. It matches Java today because payload wins for v6, but a pre-v8 table with no payload class would fall to the ordering-field guess instead of its strategy id. Was omitting the pre-v8 key intentional?

hudi-agent (commented):

Thanks for working on this! This PR replaces the old (never-populated) merge-strategy guess with an inference that mirrors Java's `inferMergingConfigsForPreV9Table`, so previously-refused v6 MOR tables now read (via payload class → commit-time ordering), and it keeps refusing tables with a custom merger. The precedence logic and the v8 version threshold both look faithful to Java, and the unit coverage is thorough. One thing worth double-checking in the inline comment: the merge-strategy-id key this path reads matches v8+ tables but not the pre-v8 tables the inference actually serves. Please take a look at any inline comments, and this should be ready for a Hudi committer or PMC member to take it from here.

crates/core/src/file_group/reader.rs
linliu-code: Agreed and fixed. Both per-slice lines in `reader.rs` are now `debug!` with a comment giving your reason ("a wide read would repeat this once per slice, and which reader served the table is not news N times over"), and the `merge mode …` lines in `resolver.rs` are `debug!` too, since `resolve_reader_context` runs per slice for a table-level constant.

Worth flagging the consequence: there is now no `info!`-level statement of which reader served a table, so an operator diagnosing a difference between the two versions has to enable debug logging. That seems like the right trade for not emitting N identical lines per read, but it's a deliberate choice rather than an oversight.

crates/core/src/file_group/reader_v2/resolver.rs
linliu-code: Fixed. Hoisted to `let ordering_desc = if has_ordering_field { "having" } else { "not having" };` and the message is now a plain template using the inline capture, matching the house style on the lines below. (The macro is also `debug!` now — see your other comment on this PR.)

crates/core/src/file_group/reader_v2/resolver.rs
linliu-code: Not intentional — you're right, and I confirmed it against the fixtures before changing anything.

Grepping every `hoodie.properties` in the corpus: `hoodie.compaction.record.merger.strategy` appears 72 times, always `eeb8d96f-…` and always with `hoodie.record.merge.mode` absent; `hoodie.record.merge.strategy.id` appears only on tables that also set `hoodie.record.merge.mode`, which return before reaching the inference. So `strategy_id` really was always empty on this path, exactly as you said.

Fixed by reading both spellings, newest first:

```rust
const RECORD_MERGE_STRATEGY_ID_KEYS: [&str; 2] = [
    "hoodie.record.merge.strategy.id",
    "hoodie.compaction.record.merger.strategy",
];
```

One thing worth recording: this changes no current answer. All 72 of those fixtures also name `hoodie.compaction.payload.class=OverwriteWithLatestAvroPayload`, and the pre-v8 branch already keeps the payload class authoritative, so payload still wins for them — which is what Java does. What the fix closes is precisely the case you named: a pre-v8 table with no payload class now uses its strategy id instead of falling through to the ordering-field guess.

linliu-code:

PR #673, opened 2026-08-07
feat(core): incremental reads, position merge, and the gold sweep that checks them

linliu-code:
Stacked on #672. GitHub cannot show that: a pull request's base branch has to
exist in the repository it targets, and these branches live in a fork, so every
PR in this line is opened against `main` and shows its predecessors' commits
too. **This PR's own change is 26 commits, +7,673/-1,041:**

<https://github.com/linliu-code/hudi-rs/compare/fix/v2-merge-mode-inference...feat/v2-gap-reporting>

The 19 commits below those come from #663-#672 and are reviewed there. #660
carries the whole line squashed, for a single-diff read of everything.

### What this PR adds

**Incremental reads on merge-on-read tables.** They did not work; the slice
live at the end of the range is now read and its rows narrowed back to the
window by commit time. A fixture whose base file was written by compaction
spans a window boundary, so admitting the file must not admit every record in
it.

**A gold parity sweep.** Every fixture, both engines, against a Spark snapshot
of the same table — rather than against values we wrote down ourselves. Each
fixture ships `gold_data/`, and the merge-on-read corpus additionally ships
`gold_options/`: one snapshot per read-option case, with a manifest naming
them. The manifest is written by the generator that produces the snapshots and
only read by the test, so a case is never spelled out twice.

**The option matrix.** A `SELECT *` snapshot only proves the engines agree when
nothing is asked of them. 183 cases now cover projection (including dropping
the record key and the ordering field, which the merge needs internally),
read-optimized, and incremental windows.

**A configurable log-read buffer.** `hoodie.memory.dfs.buffer.max.size` bounds
how much of a log file one reader holds. It was hardcoded at 16 MiB — the same
value Hudi defaults to, so nothing read differently, but setting it did nothing
and said nothing. It matters because the cost scales with concurrency, and this
crate reads four file slices at a time by default.

**Position-based merge.** `hoodie.merge.use.record.positions` reaches the
engine and does what it says. The `table_duplicate_keys` fixture is the one
place the two strategies can disagree, and it asserts they do — otherwise the
position sweep would pass even if position merge silently fell back to key
merge.

**Incremental reads ranged on completion time.** Hudi 1.x admits a commit by
its completion time, not the time it was requested, so ranging on requested
time diverged whenever a window straddled one commit's requested/completion
pair. This changes what an incremental read returns, hence the `!` on that
commit.

**Gap reporting.** `reader_v2::gaps` collects, as data, the behaviors the
engine does not reproduce. Only *silent* gaps belong there: anything already
refused says so through its error, and repeating it as a warning would train
callers to ignore warnings.

### Bugs the sweep found, and this PR fixes

- A file group's log files were placed in **directory-listing order**, so a
  log-only group split into two file slices depending on what order the
  filesystem happened to return. A read as of an instant then saw only the
  later half. Surfaced as a *wider* incremental window returning *fewer* rows.
  Reachable from the metadata table reader and both language bindings.
- `HoodieTimeline.INIT_INSTANT_TS` (`"00000000000000"`) — the canonical "from
  the beginning" incremental bound, which Spark accepts — was **rejected**. The
  epoch-millis fallback only applied to 17-character values, leaving the
  14-character sentinels matching neither branch.
- A slice was read with the base file's schema rather than the table's; a
  promoted column was reinterpreted rather than cast; a slice with no base file
  was neither discovered nor representable.

### Where it stands

The merge-on-read engine matches Hudi everywhere the corpus reaches:

```
compared 62 fixtures — legacy 50, v2 62; 7 without gold
compared 62 fixtures — v2+positions 62; 7 without gold
option cases on 23 fixtures — legacy 125, v2 183
```

The legacy reader's 12 remaining disagreements each have a named cause in the
sweep's known list — a panic on partial-update blocks, dropped base columns on
a log-only slice, collapsed duplicate keys, and others. That list is a ratchet:
a new disagreement fails the build, and one that stops happening must be
removed, which is enforced rather than asked for.


crates/core/src/table/mod.rs
```diff
@@ -955,6 +1140,53 @@ impl Table {
...
+
+        // The boundary itself, and anything after it, is wholly inside the
+        // active timeline.
+        assert!(
+            hudi_table
+                .window_predates_active_timeline(&boundary)
+                .is_none(),
+            "a window starting exactly at the boundary covers no archived commit"
+        );
+        let after = format!("{boundary}9");
```
hudi-agent, line 1183: 🤖 nit: the `"9"` suffix is a little magic here — could you add a short comment explaining that appending any character to the boundary string produces a string that sorts after it lexicographically, so the window lands inside the active timeline? Something like `// any suffix makes this sort after boundary in Hudi's lex comparison`.

hudi-agent (commented):

Thanks for working on this! This slice of the stack refactors `warn_if_window_predates_active_timeline` into an assertable `window_predates_active_timeline` helper (which reads as behavior-preserving) and adds regression tests for the incremental-window warning and for layout-v1 archived-instant enumeration. No correctness issues found. A few style/readability suggestions in the inline comments. Please take a look, and this should be ready for a Hudi committer or PMC member to take it from here. One small nit on a magic literal in the test; everything else is clean and well-commented.

cc @yihua

crates/core/src/table/mod.rs
linliu-code: Fixed. Added, close to your wording:

```rust
// Hudi compares instant times lexicographically, so appending any
// character yields a string that sorts strictly after the boundary —
// i.e. a window that opens inside the active timeline.
```

linliu-code:

PR #677, opened 2026-08-10
feat(core): select the file group reader by version, defaulting to 2

linliu-code:
Foundation for landing a second merge implementation. **3 files, +435/−2.** No read changes.

### What this establishes

`hoodie.read.file.group.reader.version` selects `2` (the default) or `1`.
`FileGroupReader::read_file_slice_from_paths` becomes the single decision point for reads that
merge — which is every eager read, and the streaming reads that have log files:

```
read_file_slice_from_paths(base, logs, options)
  ├─ metadata table            → existing reader   (always, setting ignored)
  ├─ version == 1              → version 1         (explicit escape hatch)
  ├─ version 2 cannot serve it → version 1         (reason logged)
  ├─ version == 2              → version 2
  └─ no logs / read-optimized  → base-file read
                                     │
                       all paths ────┴──→ apply_eager_options(filters, projection)
```

`Table`, DataFusion, the Python binding and the cxx bridge are untouched.

**One asymmetry worth naming.** `read_file_slice_from_paths_stream_inner` answers two cases
before it would reach the switch: a read-optimized stream, and a stream over a slice with no log
files. Both go straight to `read_base_file_stream`, so the version setting is not consulted there.
That is harmless in this PR — every capability falls back, so version 2 is never entered
anywhere — but it means a streaming read-optimized query would silently keep using the existing
reader once version 2 becomes capable. Whether the streaming path should consult the switch is a
decision worth making deliberately rather than inheriting from the order of these branches; it is
left as-is here so the seam lands without changing streaming behaviour.

### Why version 2 is already the default

It serves nothing yet, so the fall back makes the default inert — a test asserts a real
merge-on-read slice returns the same rows as an explicit version 1 read.

Making it the default only once it were capable would put the whole behaviour change in a single
commit whose diff shows nothing functional — the scariest possible shape to review. This way
that commit never exists: each capability-adding PR carries its own small, individually
revertable behaviour change, and *"v2 now serves base-file-only reads"* is a claim a reviewer can
actually evaluate.

**The consequence to hold in review:** a failure *inside* version 2 now reaches every caller,
not just one who opted in. The capability check must therefore stay conservative — when in
doubt, report unsupported. Today that risk is zero, because nothing is claimed.

version 1 remains selectable, so a caller can opt out entirely rather than relying on fall back.

### Selecting a version is a request, not a guarantee

`version_two_unsupported_reason()` runs first and names what version 2 cannot serve. Those reads are
served by the existing reader instead, so **the setting cannot turn a working read into a failing
one** — and version 2 can land incrementally behind a contract that removes one reason at a
time, rather than needing full parity before it is reachable.

Every fall back is logged. A fall back nobody can observe is indistinguishable from a reader
that is never used.

### The distinction that keeps fall back safe

The check is decided **from config before any I/O — never a catch-all on error.** A read that
fails *inside* version 2 propagates. Retrying it on the existing reader would make a bug look
like a success, make results depend on which reader happened to win, and leave differential
tests unable to see anything.

### Two cases error rather than fall back

- **`CUSTOM` record merge mode, when the read actually merges.** It needs a merger neither reader
  implements, and falling back would use the existing reader's own derivation, which drops
  deletes — wrong rows are worse than a refusal. The refusal is gated on a merge happening
  (`log_file_paths.is_empty() || is_read_optimized()`): a base-file-only read never consults a
  merger, so refusing it would make a `CUSTOM` table unreadable for queries that were always
  correct. Read by raw key (`hoodie.record.merge.mode`) through a borrowing `HudiConfigs::get_raw`;
  this crate has no typed config for it yet, and adding one belongs with the reader that acts on it.
- **A metadata table** is routed to the existing reader *before* the check, since its base files
  and log blocks are HFile. The error behind that routing is an unreachable guard, so a future
  change to the routing fails loudly rather than reaching a reader that cannot read HFile.
  Erroring outright would make metadata tables unreadable whenever the setting is applied
  globally — and table listing reads one.

### An unrecognised value is an error

Not a fall back to the default. `get_or_default` would have swallowed a typo and read with the
other version, leaving a caller convinced they had exercised the one they asked for. That is the
one outcome a switch like this must not produce.

The lookup goes through `HudiConfigs::try_get`, which already had exactly these semantics —
it parses and returns `Err` rather than falling back. Leniency stays the right default for most
configs; it is specifically wrong here.

### The rule this seam exists to make enforceable

The fall back handles what version 2 *knows* it cannot do. Nothing here defends against version 2 returning
plausible but wrong rows — and nothing at runtime can: a reader that knew its answer was wrong
would not be wrong. Only a differential comparison against Hudi's own reader catches that, and
that is a test harness, not a guard.

So the safety property is procedural, and worth agreeing before the first capability lands rather
than after:

> **A capability may only be claimed together with fixture coverage comparing its output against
> Hudi's reader.**

Two things in this PR are shaped to support it:

- **Capability is enumerated, not assumed.** The fallthrough of `version_two_unsupported_reason` reports
  *unsupported*, so a situation nobody considered falls back rather than being served by a reader
  that has never seen it. Inverting that is a one-line change with no visible symptom, which is
  why it is called out in a comment at the site.
- **The "version 2 reports itself capable" branch errors**, and its message says a capability must
  not be claimed before that coverage exists — so the requirement is in the source a contributor
  will actually hit, not only in this description.

What is still missing, and belongs with the first capability rather than here: the differential
harness itself, and a counter or read-stat making "is version 2 serving anything?" answerable in
production. A fall back nobody can observe is indistinguishable from a reader that is never used
— we have been caught by exactly that twice.

### Where the check is going: a positive capability list

The check here is a **negative** list — enumerate what version 2 cannot do, fall through to
unsupported. Safe, but the safety is a convention: the fallthrough is one line from being
inverted with no visible symptom.

The intended next step, landing with the first capability rather than here, is to make it
**positive and structural**:

```rust
enum SupportedCategory { BaseFileOnly, /* … */ }

impl SupportedCategory {
    /// Fixtures proving this category: …
    fn matches(&self, configs: &HudiConfigs, options: &ReadOptions,
               base_file_path: &str, log_file_paths: &[String]) -> bool
}

/// `None` → no category claims this read → fall back.
fn supported_category(...) -> Option<SupportedCategory>
```

version 2 serves a read only if some category claims it. Adding a variant forces every match arm to be
revisited, and a query shape nobody considered cannot be served, because no category claims it —
enforcement in the type system rather than in a comment.

It also makes the coverage rule above checkable: each variant names, in its doc comment, the
fixtures that prove it. *"A capability may only be claimed with fixture coverage"* becomes
something a reviewer verifies by reading the enum.

The two checks stay separate, and the order matters:

| | question | outcome |
|---|---|---|
| `version_two_unsupported_reason` | *should anyone serve this?* | **error** — `CUSTOM` merge mode, metadata table |
| `SupportedCategory` | *has version 2 been proven to serve this?* | **fall back** |

A `CUSTOM` table must error even if a category would otherwise match it, so refusals run first.

Two notes on shape. Categories should track the dimensions the code actually branches on —
base-file-only vs merge, log block format, query type — not table version or keygen, which the
reader does not branch on; too coarse claims too much, too fine sprawls. And the validator cannot
work from configs alone: whether a slice has log files is per-slice, not per-table, so it takes
the same inputs the dispatch already has.

**A category is an assertion, not a proof.** `BaseFileOnly` can match a base file whose schema
evolution version 2 mishandles. Categories narrow the claim; the fixture coverage verifies it. That is
why tying each variant to its fixtures matters more than the enum itself.

It is not introduced in this PR because there are zero supported categories yet — an enum with no
variants cannot be matched exhaustively and is dead code. It arrives with its first real variant,
and its first proof.

### On the name

`hoodie.read.file.group.reader.version`, with bare integers, replaces an earlier
`hoodie.read.merge.engine = legacy | v2`. Three reasons, all worth settling before release since
this becomes public API:

- **`merge.engine` collided.** Hudi already has `hoodie.datasource.merge.type`
  (`skip_merge` / `payload_combine`), which is a *semantic* choice — what merging does. This is an
  *implementation* choice — which code does it. "Engine" is also overloaded in Hudi, where it
  usually means the compute engine.
- **`legacy` was a value judgement that inverts.** Once the newer reader is the default, `legacy`
  is what you deliberately opt *into*; asking a user to set "legacy" to get a working read reads
  badly. A version number carries no such judgement.
- **A version says the older one is being retired**, not kept as an alternative. Names like
  `batch_merge` / `record_merge` would imply a permanent choice between two valid strategies,
  which is not the intent.

Bare integers match how Hudi already versions things — `hoodie.table.version`,
`hoodie.timeline.layout.version`, `hoodie.write.table.version` — and make the config an integer,
so an unparseable value errors without string matching.

### Testing

Eight tests: the default is version 2 and still falls back; version 1 stays selectable as an
escape hatch; an unrecognised value errors and names it; asking for version 2 falls back with a
reason; **a real merge-on-read slice returns the same rows under the default as under an explicit
version 1 read** — compared cell by cell, since a row count matches even when the columns or the
values do not; a metadata table ignores the setting; and `CUSTOM` errors when the read merges but
falls back with a reason when it does not.

Every capability is unimplemented in this PR, so version 2 is never entered. If it ever reports itself
capable before version 2 is wired up, the read errors loudly rather than silently doing nothing.

Full suite green: 718 core + 79 table-read + 39 datafusion. `cargo fmt --check` and clippy clean.

### DIFF
--- a/crates/core/src/file_group/reader.rs
+++ b/crates/core/src/file_group/reader.rs
@@ -49,7 +49,6 @@ use futures::stream::BoxStream;
 use futures::{StreamExt, TryFutureExt};
 use std::collections::HashMap;
 use std::convert::TryFrom;
-use std::str::FromStr;
 use std::sync::Arc;
 
 /// The reader that handles all read operations against a file group.
@@ -238,12 +237,17 @@ impl FileGroupReader {
         if self.is_metadata_table() {
             return Ok(FileGroupReaderVersion::One);
         }
+        // `try_get` rather than `get_or_default`: the latter returns the default
+        // when a value fails to parse, so a typo would silently read with the
+        // other version and leave a caller convinced they had exercised the one
+        // they asked for. Borrowing, so no copy of the config map per read.
         match self
             .hudi_configs
-            .as_options()
-            .get(HudiReadConfig::FileGroupReaderVersion.as_ref())
+            .try_get(HudiReadConfig::FileGroupReaderVersion)?
         {
-            Some(raw) => FileGroupReaderVersion::from_str(raw).map_err(CoreError::Config),
+            Some(value) => {
+                FileGroupReaderVersion::try_from(usize::from(value)).map_err(CoreError::Config)
+            }
             None => Ok(FileGroupReaderVersion::default()),
         }
     }
@@ -274,12 +278,10 @@ impl FileGroupReader {
         // consulting a merger at all, so refusing it would break reads that work
         // today over a mode they never reach.
         if !base_file_only
-            && let Some(mode) = self
-                .hudi_configs
-                .as_options()
-                // Read by raw key: this crate has no typed config for it yet, and
-                // adding one belongs with the reader that acts on it.
-                .get("hoodie.record.merge.mode")
+            // Read by raw key: this crate has no typed config for it yet, and
+            // adding one belongs with the reader that acts on it. Borrowed, so
+            // no copy of the option map per read.
+            && let Some(mode) = self.hudi_configs.get_raw("hoodie.record.merge.mode")
             && mode.eq_ignore_ascii_case("CUSTOM")
         {
             return Err(CoreError::Unsupported(
@@ -1690,7 +1692,6 @@ mod tests {
 #[cfg(test)]
 mod file_group_reader_version_tests {
     use super::*;
-    use crate::config::util::empty_options;
     use hudi_test::SampleTable;
 
     async fn reader_with(
--- a/crates/core/src/file_group/reader.rs
+++ b/crates/core/src/file_group/reader.rs
@@ -225,12 +225,12 @@ impl FileGroupReader {
     ///
     /// A metadata table is always served by version 1 whatever the
     /// setting says: its base files and log blocks are HFile, which the
-    /// merge-on-read engine has no support for. That is permanent, not
+    /// file group reader version 2 has no support for. That is permanent, not
     /// transitional.
     ///
     /// The value is read raw rather than through `get_or_default`, which falls
-    /// back to the default when a value fails to parse. A typo in the engine
-    /// name would then silently read with the other engine, leaving a caller
+    /// back to the default when a value fails to parse. A typo in the version
+    /// would then silently read with the other version, leaving a caller
     /// convinced they had exercised it — the one outcome this switch must not
     /// produce.
     fn file_group_reader_version(&self) -> Result<FileGroupReaderVersion> {
@@ -252,18 +252,18 @@ impl FileGroupReader {
         }
     }
 
-    /// Why the merge-on-read engine cannot serve this read, if it cannot.
+    /// Why file group reader version 2 cannot serve this read, if it cannot.
     ///
     /// This is a capability check, decided from config before any I/O — never a
-    /// catch-all on error. A read that fails *inside* the engine propagates:
+    /// catch-all on error. A read that fails *inside* version 2 propagates:
     /// retrying it on version 1 would make a bug look like a success,
-    /// make results depend on which engine happened to win, and leave the
+    /// make results depend on which reader happened to win, and leave the
     /// differential tests unable to see anything.
     ///
     /// Every reason here means version 1 serves the read instead, so
-    /// selecting the engine cannot turn a working read into a failing one. Each
+    /// selecting a version cannot turn a working read into a failing one. Each
     /// reason is logged, because a fallback nobody can observe is
-    /// indistinguishable from an engine that is never used.
+    /// indistinguishable from a reader that is never used.
     fn version_two_unsupported_reason(
         &self,
         options: &ReadOptions,
@@ -296,7 +296,7 @@ impl FileGroupReader {
         // loudly here rather than reaching a reader that cannot read HFile.
         if self.is_metadata_table() {
             return Err(CoreError::Unsupported(
-                "The merge-on-read engine cannot read a metadata table's HFile \
+                "File group reader version 2 cannot read a metadata table's HFile \
                  base files and log blocks"
                     .to_string(),
             ));
@@ -307,10 +307,10 @@ impl FileGroupReader {
         }
         // The fallthrough reports *unsupported*, deliberately: capability is
         // enumerated, not assumed, so a situation nobody considered falls back
-        // rather than being served by an engine that has never seen it.
+        // rather than being served by a reader that has never seen it.
         // Inverting this is a one-line change with no visible symptom, which is
         // why it is called out here.
-        Ok(Some("the merge-on-read engine is not wired up yet"))
+        Ok(Some("file group reader version 2 is not wired up yet"))
     }
 
     /// Reads a file slice from a base file and a list of log files.
@@ -346,8 +346,8 @@ impl FileGroupReader {
                     // that is a test harness. So a capability may only be added
                     // together with fixture coverage proving it.
                     return Err(CoreError::Unsupported(
-                        "The merge-on-read engine reports itself able to serve this read, \
-                         but no engine is wired up behind the switch yet. A capability must \
+                        "File group reader version 2 reports itself able to serve this read, \
+                         but nothing is wired up behind the switch yet. A capability must \
                          not be claimed here before there is fixture coverage comparing its \
                          output against Hudi's reader"
                             .to_string(),
@@ -1701,7 +1701,7 @@ mod file_group_reader_version_tests {
         FileGroupReader::new_with_options(base_url.as_ref(), options).await
     }
 
-    /// The merge-on-read engine is the default, and nothing changes for a caller
+    /// File group reader version 2 is the default, and nothing changes for a caller
     /// who sets nothing — because every capability falls back today. Making it
     /// the default only once it were capable would put the whole behaviour change
     /// in one commit; this way each capability carries its own.
@@ -1731,7 +1731,34 @@ mod file_group_reader_version_tests {
         Ok(())
     }
 
-    /// A typo must not read with the other engine. `get_or_default` would have
+    /// The guard inside the check, reached only if the dispatch's metadata
+    /// routing were ever removed. Asserted directly because the dispatch answers
+    /// metadata tables before the check runs, so no read can reach it today —
+    /// which is exactly why it must keep erroring rather than fall through to a
+    /// reader that cannot read HFile.
+    #[tokio::test]
+    async fn test_version_two_unsupported_reason_metadata_table_returns_error() -> Result<()> {
+        use crate::config::HudiConfigs;
+        use crate::config::table::HudiTableConfig;
+        use std::collections::HashMap;
+        use std::sync::Arc;
+
+        let configs = Arc::new(HudiConfigs::new([(
+            HudiTableConfig::BasePath.as_ref(),
+            super::tests::get_metadata_table_base_uri(),
+        )]));
+        let reader = FileGroupReader::new_with_overrides(configs, HashMap::new(), HashMap::new())?;
+        let err = reader
+            .version_two_unsupported_reason(&ReadOptions::new(), false)
+            .unwrap_err();
+        assert!(
+            matches!(err, CoreError::Unsupported(ref m) if m.contains("metadata table")),
+            "expected an unsupported error naming the metadata table, got {err:?}"
+        );
+        Ok(())
+    }
+
+    /// A typo must not read with the other version. `get_or_default` would have
     /// swallowed this and left the caller believing they had exercised `v2`.
     #[tokio::test]
     async fn test_file_group_reader_version_unrecognised_returns_config_error() -> Result<()> {
@@ -1752,7 +1779,7 @@ mod file_group_reader_version_tests {
         Ok(())
     }
 
-    /// Asking for the engine is a request, not a guarantee: every capability is
+    /// Asking for a version is a request, not a guarantee: every capability is
     /// unimplemented so far, so version 1 serves the read and says why.
     #[tokio::test]
     async fn test_version_two_unsupported_reason_nothing_implemented_returns_reason() -> Result<()>
@@ -1770,7 +1797,7 @@ mod file_group_reader_version_tests {
         let reason = reader.version_two_unsupported_reason(&ReadOptions::new(), false)?;
         assert!(
             reason.is_some(),
-            "with no engine wired up, every read must fall back"
+            "with nothing wired up, every read must fall back"
         );
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


crates/core/src/file_group/reader.rs
```diff
@@ -221,6 +222,88 @@ impl FileGroupReader {
...
+        // commit-time-ordered table. Wrong rows are worse than a refusal.
+        if let Some(mode) = self
+            .hudi_configs
+            .as_options()
+            // Read by raw key: this crate has no typed config for it yet, and
+            // adding one belongs with the reader that acts on it.
+            .get("hoodie.record.merge.mode")
+            && mode.eq_ignore_ascii_case("CUSTOM")
+        {
+            return Err(CoreError::Unsupported(
```
hudi-agent, line 287: 🤖 Because v2 is the default, this CUSTOM check runs on every non-metadata read, and it fires before the read-optimized/base-file-only determination. So a CoW snapshot or a read-optimized MoR read — which never merges — would now hard-error on any table declaring `hoodie.record.merge.mode=CUSTOM` (loaded verbatim into configs by `imbue_table_properties`), even though those reads returned correct data before this PR. Could the refusal move below the `base_file_only`/`is_read_optimized` determination so it only triggers when a merge actually happens?

crates/core/src/file_group/reader.rs
```diff
@@ -1568,3 +1674,163 @@ mod tests {
...
+        options: impl IntoIterator<Item = (&'static str, String)>,
+    ) -> Result<FileGroupReader> {
+        let base_url = SampleTable::V6Nonpartitioned.url_to_mor_parquet();
+        FileGroupReader::new_with_options(base_url.as_ref(), options).await
+    }
+
+    /// The merge-on-read engine is the default, and nothing changes for a caller
+    /// who sets nothing — because every capability falls back today. Making it
+    /// the default only once it were capable would put the whole behaviour change
+    /// in one commit; this way each capability carries its own.
```
hudi-agent, line 1707: 🤖 nit: the test names in this module read as prose sentences rather than the `test_<function>_<scenario>_<expected>` convention used elsewhere in the crate — could you rename them along those lines? For example, `the_default_version_is_two_and_still_falls_back` could become `test_file_group_reader_version_default_returns_two` and `an_unrecognised_version_is_an_error` could become `test_file_group_reader_version_unrecognised_value_returns_config_error`.

crates/core/src/file_group/reader.rs
```diff
@@ -221,6 +222,88 @@ impl FileGroupReader {
...
+    /// A metadata table is always served by the legacy reader whatever the
+    /// setting says: its base files and log blocks are HFile, which the
+    /// merge-on-read engine has no support for. That is permanent, not
+    /// transitional.
+    ///
+    /// The value is read raw rather than through `get_or_default`, which falls
+    /// back to the default when a value fails to parse. A typo in the engine
+    /// name would then silently read with the other engine, leaving a caller
+    /// convinced they had exercised it — the one outcome this switch must not
+    /// produce.
```
yihua, line 235: nit: should `get_or_default` be fixed, or should a new config-fetching method be added to throw an error upon misconfiguration?

crates/core/src/file_group/reader.rs
```diff
@@ -221,6 +222,88 @@ impl FileGroupReader {
...
+    fn file_group_reader_version(&self) -> Result<FileGroupReaderVersion> {
+        if self.is_metadata_table() {
+            return Ok(FileGroupReaderVersion::One);
+        }
+        match self
+            .hudi_configs
+            .as_options()
```
yihua: `HudiConfigs::as_options` is `self.raw_options.as_ref().clone()`.  Consider avoiding cloning the full configs for the file group reader version fetch.

crates/core/src/file_group/reader.rs
```diff
@@ -244,6 +327,29 @@ impl FileGroupReader {
...
+                    return Err(CoreError::Unsupported(
+                        "The merge-on-read engine reports itself able to serve this read, \
+                         but no engine is wired up behind the switch yet. A capability must \
+                         not be claimed here before there is fixture coverage comparing its \
+                         output against Hudi's reader"
+                            .to_string(),
+                    ));
+                }
+                Some(reason) => {
+                    log::debug!("reading '{base_file_path}' with the legacy engine: {reason}")
```
yihua: nit: use `file group reader version 1` instead of `legacy engine` to be clear

crates/core/src/file_group/reader.rs
```diff
@@ -221,6 +222,88 @@ impl FileGroupReader {
             .await
     }
 
+    /// Which merge implementation serves this read.
+    ///
+    /// A metadata table is always served by the legacy reader whatever the
```
yihua: nit: make the docs and code consistent, i.e., use `reader version 1` instead of `legacy reader` which can be confusing

crates/core/src/file_group/reader.rs
```diff
@@ -1568,3 +1674,163 @@ mod tests {
...
+                )],
+                None => Vec::new(),
+            };
+            let reader = FileGroupReader::new_with_options(base_url.as_ref(), options).await?;
+            let mut rows = 0;
+            for slice in &slices {
+                rows += reader
+                    .read_file_slice(slice, &ReadOptions::new())
+                    .await?
+                    .num_rows();
```
yihua: Validate row and column values beside count to be stricter.

crates/core/src/file_group/reader.rs
```diff
@@ -244,6 +327,29 @@ impl FileGroupReader {
             .collect();
         let base_file_only = log_file_paths.is_empty() || options.is_read_optimized()?;
```
yihua, line 337: To clarify, the base file-only case is still going through the existing base file reader (not the file group reader), which does not require record merging, correct?

crates/core/src/file_group/reader.rs
yihua: +1

crates/core/src/file_group/reader.rs
yihua: +1

yihua (commented):
@linliu-code Thanks for putting this up.  I leave one major comment around the CUSTOM merge mode regression and a few nits.

crates/core/src/file_group/reader.rs
yihua: Let's add a regression test around this.

crates/core/src/file_group/reader.rs
linliu-code: Done in 42e1fab — the rename left "legacy reader"/"legacy engine" behind in the docs, which is exactly the confusion the rename was meant to remove. There are now zero occurrences of "legacy" in this file; it reads "version 1" throughout.

crates/core/src/file_group/reader.rs
linliu-code: Done in 42e1fab. The log line now reads `reading '<path>' with file group reader version 1: <reason>` — worth getting right since this is the message a caller actually sees when a read falls back.

crates/core/src/file_group/reader.rs
linliu-code: Done in 42e1fab. Renamed all six to `test_<function>_<scenario>_<expected>` per crates/CLAUDE.md, e.g. `test_file_group_reader_version_unset_returns_two`, `test_file_group_reader_version_unrecognised_returns_config_error`, `test_read_file_slice_from_paths_default_version_matches_version_one`.

One of them also asserted the fall back on top of the resolved version, which is what the next test is for — each now asserts only what its name claims.

crates/core/src/file_group/reader.rs
linliu-code: Correct — `read_base_file_eager`, the base file reader, with no merging. Neither file group reader is involved, and the version setting does not reach that branch.

Added a comment at the branch in 42e1fab saying so, since it is also the reasoning behind the fix for your other point on line 278: a read with nothing to merge should not be refused for a merge-mode it never consults.

crates/core/src/file_group/reader.rs
linliu-code: Confirmed and fixed in 74ef65d — you are right, and it is worse than a nit because version 2 is the default, so nobody opted in.

The refusal now takes `base_file_only` and applies only when a merge actually happens. The reason for refusing is that falling back would merge with version 1's derivation, which drops deletes — that reasoning simply does not apply to a read that never consults a merger, so a copy-on-write or read-optimized read falls back like any other unimplemented capability.

I kept the dispatch where it is rather than moving it below the `base_file_only` branch, because version 2 is meant to serve base-file-only reads later; it was the refusal that was mis-scoped, not the dispatch.

linliu-code: Added in 74ef65d — two tests, one either side of the gate:

- `..._custom_merge_mode_without_merge_returns_reason` — a CUSTOM table with nothing to merge falls back instead of erroring, checked for both the no-log-files and read-optimized routes.
- `..._custom_merge_mode_with_merge_returns_error` — the same table is still refused once merging is involved, and the error names why.

Checked the first one is not vacuous: removing the `!base_file_only` gate makes it fail, and restoring it makes it pass.

crates/core/src/file_group/reader.rs
linliu-code: Done in 74ef65d. It now renders every cell of every row (and the column names) and compares those, rather than summing `num_rows()` — a count would match even if the columns or the values differed.

crates/core/src/file_group/reader.rs
linliu-code: Done in ef0a708 — and it turned out neither fixing `get_or_default` nor adding a new accessor was needed: `HudiConfigs::try_get` already has exactly these semantics. Its doc says it "Returns `Err` on parse failures (e.g. `"yes"` for a bool config) instead of silently falling back to the default." I had hand-rolled a worse copy of a method the crate has had all along.

So the lookup is now:

```rust
match self.hudi_configs.try_get(HudiReadConfig::FileGroupReaderVersion)? {
    Some(value) => FileGroupReaderVersion::try_from(usize::from(value)).map_err(CoreError::Config),
    None => Ok(FileGroupReaderVersion::default()),
}
```

with a new `TryFrom<usize>` on the enum so the reader consumes the already-parsed integer rather than re-parsing the string. `get_or_default` stays as-is — leniency is the right default for most configs; it's specifically wrong here, where silently falling back would run the *other* reader while the caller believes they exercised the one they asked for. `test_file_group_reader_version_unrecognised_returns_config_error` pins that.

crates/core/src/file_group/reader.rs
linliu-code: Fixed in ef0a708. `as_options()` clones the entire option map, so both lookups were copying it on every read. The version lookup now goes through `try_get`, which borrows.

The merge-mode lookup still reads a raw key — this crate has no typed config for `hoodie.record.merge.mode` yet, and adding one belongs with the reader that acts on it rather than with this seam. But it no longer clones: I added a borrowing `HudiConfigs::get_raw` next to the existing `contains`, documented as the escape hatch for keys with no typed config, pointing at `try_get` where one exists.

The one remaining `as_options()` in this file is in the constructor, which runs once per reader rather than once per read.

yihua (approved):
LGTM.  Could you fix the test coverage reported by Codecov?

linliu-code:
> LGTM. Could you fix the test coverage reported by Codecov?

Addressed. but need to another run to confirm.

PR #678, opened 2026-08-15
fix(core): gate the log-block scan on instant state

linliu-code:
## Why

`forward_scan_pass1` has five gates, and Gate 3 checks whether a log block's instant actually completed. It never ran: `completion_gate_inputs` was `None` at every construction site this crate reaches, so the gate was inert while its docs read as if it were wired. Raised on #660 ([resolver.rs:89](https://github.com/apache/hudi-rs/pull/660#discussion_r3788329725), [again](https://github.com/apache/hudi-rs/pull/660#discussion_r3787843003)).

The gap is the straddling case. A log file is admitted to a file slice on its own committed instant, but the blocks *inside* it carry their own instants. A writer that opens an instant and never finishes leaves blocks whose instant sorts below the latest committed one — so they clear the future gate and the range gate, and nothing asks whether that instant completed. Those rows merge as if committed.

Not a regression: reader version 1 has no committed-set check either. But version 2 is the default now, `Table` already holds the timeline the gate needs, and the fix is cheap.

## What changed

`Timeline::new_from_storage` already lists every active instant in every state to compute the archival boundary, then discards all but the completed ones. It now keeps the rest as `pending_instants`, and the gate inputs travel `Table` → `FileGroupReader` → adapter the same way the data schema already does, shared as an `Arc` since a read hands them to every file slice.

## Scoped to table version < 8, matching Java

`Table::completion_gate_inputs` supplies the sets only below table version 8, mirroring `BaseHoodieLogRecordReader`'s `tableVersion.lesserThan(EIGHT)`. The reason is in how log files are named:

- **Below version 8** a log file's name carries the *base* instant, not the delta commit that wrote it. No file-level check can attribute the file to a commit, so the per-block gate is the only place the question can be asked.
- **From version 8** each log file carries its own delta commit, and the writer never appends to an existing file: `HoodieLogFormatWriter.getOutputStream()` rolls over to a new version when the path already exists, then creates it fresh. Every block in a file therefore belongs to the commit its name records, so dropping the file when the slice is built is complete rather than merely usually sufficient, and the per-block gate would only re-ask a settled question.

Java scoped it the same way and did so deliberately, in HUDI-9567 (`3369b09fd8fc`), which added the version condition to a check that had been unconditional. The two fixtures show the split directly: disarming the gate fails `table_uncommitted_log_v6` on reader version 2 and leaves `table_uncommitted_log_v9` passing, because v9 reaches the same answer by the file-level route.

Stated plainly, the condition is parity and avoided work rather than an observable fix: arming the gate on every version also passes the whole sweep, so nothing in the corpus distinguishes the two. It is scoped this way so a check that can never fire on v8+ does not later read as load-bearing.

A caller with no timeline at all (the cxx bridge) passes `None` and the gate stays off, as before.

An upgraded version-6 table might look like it breaks that reasoning, since its old log files keep base-instant names. It doesn't: the writer's upgrade rolls back failed writes and compacts every log file into a new base file before bumping the version, and where the pending commit has completed commits after it, rollback is refused and the upgrade aborts. Both paths were run on Spark 3.5.3 with Hudi 1.2.0-SNAPSHOT to confirm it.

## Tests

Two fixtures carry a delta commit that wrote log blocks and never completed — generated normally, then doctored by deleting the *middle* commit's completed timeline file. The middle one matters: an orphan at the end sorts above the latest committed instant, so the future-block gate discards it and Gate 3 is never reached. The first attempt at this fixture orphaned the last commit and passed with the gate disarmed.

Both table versions are present because they exclude the blocks by different routes — v6 through the per-block gate, v9 through the file slice — so the v9 fixture is what shows the redundancy rather than assuming it. `gold_data` is Spark reading the doctored table, so the expectation is Hudi's own answer.

Both also ship a `gold_options` manifest, taking the option matrix from 183 cases to 197. Version 6 gets six of the eight cases: Hudi refuses an incremental read across this table (`Found hollow commit`), so no oracle exists for those two and the generator records the skip. Version 9 has no hole in its timeline and takes all eight.

Unit tests cover the gate itself — the straddling instant, the archived boundary, composition with an incremental window, and the pending/completed subtraction.

Everything load-bearing is mutation-checked: cutting the wiring, disarming the gate, and dropping the subtraction each fail the suite, and disarming the gate additionally fails four of the new option cases on the v6 fixture.

## One correctness trap, called out because it fails silently

The timeline lists an instant **once per state file**, so a completed instant also appears as requested and inflight. Partitioning on state alone marks every committed instant pending — the gate then rejects the entire timeline and reads degrade to base-file-only with **no error at all**. The first attempt did exactly this; the gold sweep caught it. Pending is therefore the non-completed rows *minus* the completed set, pinned by a regression test.

## Behaviour changes

- A merge-on-read table below version 8 with an inflight writer appending to an already-sliced log file now excludes those uncommitted blocks instead of merging them. That is the fix.
- **Readers now disagree on such a table.** Version 1 still admits the uncommitted blocks; version 2 excludes them. This is recorded in the gold sweep's `KNOWN` list, scoped to the v6 fixture, with the reasoning and what would reopen it.
- No config surface changes.

## Verification

`make check-rust` clean · `cargo clippy -p hudi-core --lib --no-default-features -D warnings` clean · 13/13 workspace targets, 1305 lib tests · gold parity sweep green across 64 fixtures × 2 reader versions, 197 option cases · 70 Python tests against a freshly built wheel · cxx bridge builds clean.

## Known follow-ups, deliberately not in this PR

- `inflight_instants` cannot change a decision for this crate's producer, since pending is subtracted from completed and the archived boundary spans pending instants. Kept for parity with Java's two-set check and documented on the field; removing it is a standalone simplification.
- Incremental reads do not detect hollow commits, where Hudi's default is to fail. That is timeline-window resolution, a layer above this gate.


linliu-code:
Pushed `e2328bb`, a self-review catch on the first commit.

Wiring the gate to `Table` made its own docs wrong: they described the producer as "the FFI bridge when the planner enabled the gate" and scoped the gate to "table version < 8 snapshot reads" — the reader-context comment yihua flagged on #660 ([reader_context.rs:126](https://github.com/apache/hudi-rs/pull/660#discussion_r3787843187)), now in the one PR that was meant to resolve it.

The version scoping also stopped matching the code. The wiring populates the gate at both reader-construction sites unconditionally, so it applies to incremental reads too. That is intended, not an oversight: Gate 3 only ever *subtracts*, so it cannot admit a block Gate 4 rejected, and a pending instant has no business in an incremental window either.

That was an argument, so it is now a test — `test_pass1_gate3_and_the_incremental_window_only_ever_subtract` pins both directions (pending-inside-window excluded, committed-outside-window still excluded). Mutation-checked by implementing the alternative — skip the gate when a range is set — under which it is the **only** failing test in the suite, so nothing else was pinning this.

crates/core/tests/gold_parity_tests.rs
```diff
@@ -502,6 +504,26 @@ impl Known {
...
+    // default and version 1 is on its way out. Two things would reopen that: a
+    // version 1 read reachable without asking for it, or version 1 outliving the
+    // migration. Fixing it means giving `LogFileScanner` the instant state its
+    // `scan` has no notion of, which reaches the metadata table reader too.
+    Known {
+        fixture: "table_uncommitted_log_v6",
+        scope: CaseScope::Any,
+        reader_version: "1",
+        reason: "version 1 admits blocks from an instant that never completed",
+    },
```
yihua, line 526: non-blocking: `CaseScope::Any` here covers `key_only` and `read_optimized`, which v1 passes on this fixture (the orphan is an update, so the key set is unchanged, and read-optimized never touches the log). Per the `Any` scope's own doc, would it be worth naming `SelectStar` plus the four cases that actually fail, so a later v1 regression on the passing two still trips the ratchet?

yihua (commented):
LGTM overall

yihua (approved):
LGTM overall

PR #679, opened 2026-08-15
fix(core): fold partial updates in both ordering directions

linliu-code:
## Description

A KEEP_VALUES partial update that lost the ordering comparison had its columns dropped. The branch computed the union of the incoming and prior records, then handed it to the merger, which returned the prior unchanged whenever the prior won, so the union was computed and discarded. A key whose higher-ordering update arrived first lost every column unique to the later, lower-ordering update.

Java folds in both directions: when the older record wins, `DefaultSparkRecordMerger.partialMerge` passes the loser as the base and overlays the winner on top, and `SparkRecordMergingUtils.mergePartialRecords` builds the result with the overlay argument's ordering value. So the union takes the winner's columns on any overlap and the winner's ordering value, and the loser contributes only the columns the winner omits. Carrying the loser's ordering value instead would be the more damaging bug: the union would sit at the loser's position and lose to any later update above it, including the base row.

The fold now runs inside the `merge_in_place` closure, which also drops the extra `get` probe this branch was doing on top of the single-probe merge. COMMIT_TIME_ORDERING is unchanged: the incoming record is always the overlay winner, matching `OverwriteWithLatestSparkRecordMerger`.

Follow-up to #660 ([discussion](https://github.com/apache/hudi-rs/pull/660#discussion_r3788329721)).

- [ ] N/A
- [ ] Manual tests

Two unit tests in `key_based.rs` covering out-of-order partials, both mutation-checked against the pre-fix behavior: discarding the union when the prior wins returns the loser's column as unset, and carrying the incoming's ordering value makes the whole update lose to the base row. The gold parity sweep and the `--no-default-features` leg are green.


linliu-code:
Composed this branch with the async stack locally and ran gold parity, so a reviewer does not have to work out whether the two collide.

**Setup.** Cherry-picked this branch's three commits onto the stack tip (#689, `a744c12`, which sits above #688 -> #687 -> #686 -> #682 -> #678). All three applied with **zero conflicts**, including the commit that rewrites the same `merge_in_place` block in `key_based.rs` that #687 and #688 restructured.

**Results on the composed tree:**

| Check | Result |
|---|---|
| `cargo test -p hudi-core` | 1460 passed, 0 failed |
| `every_fixture_matches_hudi_on_both_reader_versions` | 65 fixtures, version 2 passes 65/65 |
| `every_fixture_matches_hudi_when_merging_by_record_position` | pass |
| `clippy -p hudi-core --lib --no-default-features -- -D warnings` | clean |
| `cargo test -p hudi-core --lib --no-default-features` | 1313 passed, 0 failed |

**Mutation check.** A clean cherry-pick plus a green sweep proves nothing on its own, so the composed tree was mutated: forcing `new_wins = true` reintroduces the pre-fix behaviour where the incoming record always wins the overlay and carries its own ordering value. That is caught by three unit tests in `key_based.rs` and by **both** gold parity sweeps, on the fixture this branch adds:

```
table_partial_update_event_time [MorAvro] reader version '2': row differs:
    actual  k1 | 200 | a-high | b-low
    gold    k1 | 300 | a-high | b-low
```

The unit test names the failure mode the description calls out as the more damaging one: `ts=5 loses to a union whose winner sat at ts=9, not to the ts=2 loser it folded`. The fix was then restored and the suite re-run green.

**What this does not cover.** The replay was this branch on top of the stack. If this lands first and the stack rebases onto it, that is a different replay, though very likely equivalent. And the auto-merge succeeded because neither side moved the same lines, so if #687 or #688 change in review this needs re-running.

Local verification only; no CI has run on this branch.

yihua (approved):
LGTM.  Thanks for contributing to reader-side fixes!

PR #680, opened 2026-08-15
test(core): cover event-time ordering losing in the gold corpus

linliu-code:
## Description

Nothing in the gold corpus made a log write lose on ordering value. Every event-time fixture either raises the ordering value on an update or leaves it alone, and a tie goes to the newer record, so the whole sweep (62 fixtures and their option cases, on both reader versions) passed unchanged when event-time ordering degraded to last-writer-wins. That semantics was asserted only by unit tests. The delete-ordering fixtures do not cover it either: they are all COMMIT_TIME_ORDERING, where the delete always applies.

This adds one v9 MOR fixture carrying both shapes in both directions: an update below the live ordering value and one above it, and a delete below and one above. A delete record carries the ordering value of the source row it matched, so `MERGE INTO` is what places a delete below the live row; a plain `DELETE` takes the live row's own value and ties. Generated with Spark 3.5.3 and Hudi 1.2.0-SNAPSHOT, the same way the rest of the corpus was, so `gold_data` is Hudi's own answer rather than ours.

Follow-up to #660 ([discussion](https://github.com/apache/hudi-rs/pull/660#discussion_r3788329728)).

- [ ] N/A
- [ ] Manual tests

The fixture joins the existing sweep and passes on both reader versions; the `KNOWN` list is not touched. Checked by mutation, each reverted after: dropping the ordering comparison in the version-2 kernel, applying a stale delete, and dropping the ordering column from the version-1 sort each fail this fixture and nothing else in the corpus.


crates/test/data/quickstart_trips_table/mor/avro/table_event_time_stale.sql
```diff
@@ -0,0 +1,69 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+-- Event-time ordering where a log update LOSES to the base row.
+--
+-- Generated with Spark 3.5.3 and Hudi 1.2.0-SNAPSHOT. gold_data is
```
linliu-code, line 22: The provenance pins Hudi 1.2.0-SNAPSHOT, same as gold_options.scala. No change needed here, but is it worth regenerating the SNAPSHOT-derived golds once 1.2.0 releases, so we are not pinning pre-release behavior long term?

linliu-code:
Thanks for working on this. This adds the first fixture where a log write loses on ordering value, which closes a real hole: the corpus previously passed intact even with event-time ordering degraded to last-writer-wins.

Checked that it passes on both reader versions with the known-disagreements list untouched, so nothing was ratcheted to make it pass, and that MOR-only is correct here since a copy-on-write snapshot has no log files to merge. One non-blocking question inline about the SNAPSHOT-derived gold.

One review pass, not an approval.

crates/test/data/quickstart_trips_table/mor/avro/table_event_time_stale.sql
linliu-code: Agreed, and leaving it as-is for now since it matches how the rest of the corpus was produced. The regeneration is worth doing once 1.2.0 ships, across every SNAPSHOT-derived gold rather than just this fixture, so it is better as its own pass than folded in here.

yihua (approved):
LGTM

PR #681, opened 2026-08-15
chore: let consumers opt out of the merge map's on-disk tier

linliu-code:
## Upgrade note (for the release notes)

`default-features = false` on `hudi` changes meaning in this release. It was previously a
no-op, because `hudi` declared no features and callers still reached the spill tier through
`hudi-core`. It now drops that tier, so a merge exceeding `hoodie.memory.merge.max.size`
returns a `CoreError::Unsupported` naming the config instead of spilling to disk. Anyone
already setting that flag defensively is affected; anyone on defaults is not.

## Description

`spill-rocksdb` exists so a consumer need not put libclang and a C++ toolchain in front of a tier that only engages above `hoodie.memory.merge.max.size`, but nothing could actually opt out: `hudi` and `hudi-datafusion` both took `hudi-core` with its defaults and forwarded nothing, so `rocksdb` reached every consumer through the umbrella crate.

Both now take `hudi-core` with `default-features = false` and forward the feature, still default-on. Forwarding from `hudi-datafusion` as well is what makes the opt-out hold: a consumer can only opt out of a feature its direct dependency exposes, and unification through that crate would otherwise return the tier whatever `hudi` asked for. The Python wheel and the cxx bridge keep the tier, since dropping it makes a merge past the budget fail instead of spilling. Note that `default-features = false` on `hudi` was previously a no-op and now genuinely removes the tier.

`AGENTS.md` described the old intent as fact and now describes this. Also cross-references the two delete-record schemas: `HoodieDeleteRecordList` is the wire schema and `HoodieDeleteRecord` its post-unwrap companion, and the singular helpers over the latter are reachable only from that module's own tests while the live decode goes through their plural namesakes, which is what made them read as dead.

Follow-up to #660 ([discussion](https://github.com/apache/hudi-rs/pull/660#discussion_r3781318368)).

- [ ] N/A
- [x] Manual tests

The `no-spill-backend` CI job now covers the umbrella crate too, where the opt-out actually happens. It builds it with `--no-default-features`, with and without the `datafusion` feature, and then asserts on `cargo tree -i rocksdb` that the tier is absent from both graphs. The build steps prove the opt-out compiles; the assertion proves it drops the dependency, and unlike the rest of that job it does not rest on the runner lacking libclang, so it keeps working if a C++ toolchain is ever added there. Checked by mutation: restoring default features on the umbrella crate's `hudi-core` dependency makes the assertion exit 1.

Locally verified by the dependency graph as well: `cargo tree -p hudi -i rocksdb` finds `rocksdb` under the defaults and finds nothing under `--no-default-features`, both with and without the `datafusion` feature. `cargo check --workspace --all-targets` and the `hudi-core` suites are green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,15 @@ jobs:
         run: cargo clippy -p hudi-core --lib --no-default-features -- -D warnings
       - name: Unit tests without the spill backend
         run: cargo test -p hudi-core --lib --no-default-features --no-fail-fast
+      # The umbrella crate is where a consumer actually opts out, and it reaches
+      # hudi-core twice: directly, and through hudi-datafusion. Either edge taking
+      # default features hands the tier back without the steps above noticing,
+      # since they only build hudi-core. Both shapes are built here, where there
+      # is no libclang to build rocksdb with.
+      - name: Check the umbrella crate without the spill backend
+        run: cargo check -p hudi --no-default-features
+      - name: Check the umbrella crate without the spill backend, with datafusion
+        run: cargo check -p hudi --no-default-features --features datafusion
 
   rust-tests-windows:
     runs-on: windows-2025
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,9 +119,36 @@ jobs:
       # so it keeps working if this job ever gains a C++ toolchain.
       - name: Assert the opt-out drops rocksdb from the dependency graph
         run: |
+          # A graph assertion has two ways to look like it passed: the tier
+          # really is gone, or cargo tree never ran. `-i` exits non-zero with an
+          # empty stdout in BOTH cases, so neither the output nor the status
+          # alone tells them apart, and discarding stderr makes a broken command
+          # indistinguishable from a working opt-out. Each check below therefore
+          # pins the specific reason rather than inferring from emptiness.
+
+          # Under defaults the tier must be reachable. If this ever stops
+          # holding, the absence checks below are asserting nothing.
+          for args in "" "--features datafusion"; do
+            if ! out=$(cargo tree -p hudi $args -i rocksdb 2>&1) \
+               || ! printf '%s\n' "$out" | grep -q '^rocksdb'; then
+              echo "cargo tree did not find rocksdb under default features: '$args'"
+              echo "so the absence checks below cannot tell an opt-out from a broken command"
+              printf '%s\n' "$out"
+              exit 1
+            fi
+          done
+
+          # With the opt-out it must be gone — and gone because the package is
+          # absent from the graph, which cargo says in those words, not because
+          # the command failed for some other reason.
           for args in "--no-default-features" "--no-default-features --features datafusion"; do
-            if cargo tree -p hudi $args -i rocksdb 2>/dev/null | grep -q '^rocksdb'; then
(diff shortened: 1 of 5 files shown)


crates/hudi/Cargo.toml
```diff
@@ -28,10 +28,21 @@ homepage.workspace = true
 repository.workspace = true
 
 [dependencies]
-hudi-core = { version = "0.5.0-dev", path = "../core" }
-hudi-datafusion = { version = "0.5.0-dev", path = "../datafusion", optional = true }
+hudi-core = { version = "0.5.0-dev", path = "../core", default-features = false }
+hudi-datafusion = { version = "0.5.0-dev", path = "../datafusion", optional = true, default-features = false }
 
 [features]
+default = ["spill-rocksdb"]
```
linliu-code, line 35: Since `hudi` had no features before this, `default-features = false` was previously a no-op and callers still got the spill tier via hudi-core. Now it drops the tier, so a merge past `hoodie.memory.merge.max.size` errors instead of spilling. Worth a line in the 0.5.0 changelog for anyone already setting that flag?

.github/workflows/ci.yml
```diff
@@ -105,6 +105,15 @@ jobs:
         run: cargo clippy -p hudi-core --lib --no-default-features -- -D warnings
       - name: Unit tests without the spill backend
         run: cargo test -p hudi-core --lib --no-default-features --no-fail-fast
+      # The umbrella crate is where a consumer actually opts out, and it reaches
+      # hudi-core twice: directly, and through hudi-datafusion. Either edge taking
+      # default features hands the tier back without the steps above noticing,
+      # since they only build hudi-core. Both shapes are built here, where there
+      # is no libclang to build rocksdb with.
+      - name: Check the umbrella crate without the spill backend
```
linliu-code, line 113: These rely on rocksdb's build script failing without libclang, same as the hudi-core steps above. Might be worth confirming on the first run that a deliberately re-coupled build does fail here, so we know the gate bites rather than assuming it.

linliu-code:
Thanks for working on this. This makes the `spill-rocksdb` opt-out actually reachable, since the umbrella crate was handing the tier back regardless of what a consumer asked for.

Verified the whole feature matrix with `cargo tree -i rocksdb`: defaults have it, `--features datafusion` has it via both edges, and both `--no-default-features` shapes drop it. `hudi-cpp` and `hudi-python` still compile and still carry the tier, so the wheel and bridge are unaffected.

Two questions inline rather than objections. Net-positive as it stands. One review pass, not an approval.

crates/hudi/Cargo.toml
linliu-code: Good point, and it turned out the changelog cannot carry it: that file is generated per release from PR titles, and this title would not convey the behavior change. Added an explicit "Upgrade note" section at the top of the PR description instead, so whoever cuts 0.5.0 has the wording to lift. It only affects consumers already passing `default-features = false`; anyone on defaults is unchanged.

.github/workflows/ci.yml
linliu-code: Fair, that was an assumption about the environment rather than a statement about the graph. Added a step that asserts directly on `cargo tree -i rocksdb` for both opt-out shapes, which does not depend on libclang being absent and keeps working if this job ever gains a C++ toolchain. Mutation-checked: restoring default features on the umbrella crate makes it exit 1.

yihua (approved):
LGTM

Issue #685, opened 2026-08-18
Make the base+log merge a Stream instead of a sync Iterator


linliu-code:
## Problem

The merge is the second reason a read crosses the sync/async boundary. Its base input is a sync `Box<dyn arrow_array::RecordBatchReader + Send>` and its output is a sync `Iterator`, so the async parquet stream has to be converted into a sync reader on the way in, and the merged result has to be converted back into a `Stream` on the way out.

Two concrete costs:

- `ParquetSyncReader` iterates the async parquet stream with `block_on(stream.next())`, which is why `make_base_file_source(streaming: bool)` returns values with two different safety contracts behind one type: the streaming branch "MUST be [consumed] in sync context", the eager branch is "safe to consume from async callers".
- `open_blocking_stream` runs the sync merge iterator on `spawn_blocking` behind a depth-1 channel to present it as a `Stream`.

Part of #684, which removes the other reason (the log read). Neither ticket alone removes all three crossings.

## Why this should be tractable

The merge is already shaped like a streaming hash join, so this is a change of interface rather than of algorithm:

- **Build**: read every admitted log block and insert its records into the merge map, folding duplicates. Necessarily complete before any output, since a base row cannot be emitted until it is known whether a log record supersedes it.
- **Probe**: pull base row groups one at a time; `merge_one_base_batch_kernel` walks each batch and `records.remove(key)` performs the join.
- **Drain**: after the base is exhausted, emit the log-only records that never matched a base row.

Build-then-probe-then-drain maps directly onto a `Stream`: await the build once, then yield one merged batch per base batch, then yield the drain in bounded chunks. The base side can stay the async parquet stream throughout, and `FileGroupMergeIterator` becomes a `Stream` impl rather than an `Iterator`.

## Depends on #684

The build phase is where log reads happen, so it can only be genuinely async once the log read is. Doing this first would mean a `Stream` whose build phase still blocks on the bridge, which trades one crossing for another rather than removing it.

## Caveats to design around

**The spill tier is synchronous.** `RocksDbDiskMap::get`, `remove` and `put` are plain blocking calls, so probing a merge map that has spilled does blocking disk I/O. Inside `poll_next` that blocks the executor thread. It is local disk rather than network, and it only happens past `hoodie.memory.merge.max.size`, so it may be acceptable, but the alternative is `spawn_blocking` per batch when the map has spilled. This needs a decision rather than an accident.

[...]

PR #686, opened 2026-08-18
perf(core): read log files asynchronously in batches, deleting the sync bridge

### LINKED ISSUE #684: Make the log-file read fully async and drop the sync/async bridge
## Problem

A single file-slice read crosses the sync/async boundary three times, and the log read is what forces all three.

| Layer | Model |
|---|---|
| Public API (`Table`, `FileGroupReader`) | async |
| Base file read | async (`ParquetRecordBatchStreamBuilder<ParquetObjectReader>`) |
| **Log file read** | **sync** (`Read + Seek` over `StorageReader`) |
| Merge | sync (takes `Box<dyn arrow_array::RecordBatchReader + Send>`) |

The crossings are: the async parquet stream wrapped in a sync reader that iterates via `block_on`; each sync log read spawning onto `OBJECT_STORE_RUNTIME` and waiting on a `sync_channel(1)`; the sync merge re-exposed as a `Stream` through `spawn_blocking` behind a depth-1 channel.

The cost is not the plumbing, it is the **non-local contracts each crossing creates**, none of which the compiler can check:

- `get_range_blocking` needs three separate verdicts about where it may be called from: a blocking-pool or plain thread is fine; an async task on another runtime's worker works but starves that worker and "remains the caller's obligation"; an `OBJECT_STORE_RUNTIME` worker is a genuine deadlock, refused by a thread-name check. `in_object_store_runtime()`, the named worker threads, and that whole body of reasoning exist only because sync code is reachable from async code.
- `make_base_file_source(streaming: bool)` returns `Box<dyn RecordBatchReader + Send>` in both branches, but one is a lazy reader that `block_on`s during iteration ("caller MUST be in sync context") and the other is fully drained ("safe to consume from async callers"). Same type, different safety contract, selected by a bool.

## Why this is now tractable

#682 already converted the log path from a cursor to **locate-then-ranged-fetch**, which is the shape parquet uses: `LogBlockContentLocation { content_position, content_length }` plus `LogBlockFetcher::read_content(offset, length)`. Only two sync I/O sites remain: the header walk, and the per-block content fetch.

`StorageReader` also has **exactly one consumer** (`file_group/log_file/reader.rs`). Nothing else in the crate uses the sync cursor, so this work deletes the bridge outright rather than removing one of several callers.

## Scope

The log read only. Making the merge a `Stream` is the other half and is tracked separately, because the two are independently shippable and have very different risk profiles.

## Proposed slices, each shippable alone

**1. Batch the content fetch.** After the gates decide, every admitted block's `(offset, length)` is known, so the fetch can be one `get_ranges(...)` instead of one serial round trip per block. Nothing in the crate currently calls `get_ranges` or `get_byte_ranges`, so every admitted block is its own round trip today, issued serially from Pass 3. On object storage that is the dominant cost.

**Caveat that must shape the design:** this pulls directly against #682, whose justification was peak memory (+200 MB to +40 MB, measured by `memory_bench`, by holding one block's content at a time). Fetching every admitted block at once gives that back. The resolution is bounded batching, coalescing within a window of K blocks and dropping each group after decode, which keeps most of the round-trip saving with a bounded peak. `hoodie.memory.dfs.buffer.max.size` already exists and means something adjacent, so it is the natural knob.

**2. Async the content fetch.** `read_content` becomes `get_range().await`. Contained, and covers the bulk of the bytes.

**3. Async the header walk.** Replace the `Read + Seek` cursor with a buffered async cursor: fetch a window, parse what it holds, advance. With this done, `get_range_blocking`, `OBJECT_STORE_RUNTIME`, `OBJECT_STORE_RUNTIME_WORKERS`, `OBJECT_STORE_RUNTIME_THREAD_NAME`, `in_object_store_runtime()` and the three call-site rules can all be deleted.

## Known risk

The walk is not a pure forward parse. `is_block_corrupted` recovers from a corrupt block by scanning forward for the next `MAGIC` over a 1 MiB window. That is the one place the parse backtracks, and it is the most likely thing to make slice 3 uglier than it looks. Worth prototyping that path first rather than last.

## Validation

The gold parity sweep is the oracle: output must stay byte-identical across all fixtures and both reader versions. `assert_lazy_matches_eager` already pins "reading lazily equals reading eagerly", which is exactly the invariant these slices must preserve. `memory_bench` measures slice 1's memory trade directly, since it already compares eager against lazy. Mutation-check each load-bearing change rather than trusting a green run.

## Non-goals

The bindings still need a runtime of their own: Python's process-wide one and the cxx per-reader `current_thread` one do not go away because the core became async. And `spawn_blocking` may still be wanted for decoding a large block; this removes it for **I/O**, which is the part that is currently misplaced.

linliu-code:
> **Stacked on #682, which is stacked on #678. Review those first.** This targets `main`, so it shows their 11 commits alongside its own 4. The delta that belongs to this PR is the last five commits: 9 files, +1102/-450.

Closes #684: slices 1, 2 and 3 are its whole scope, and together they make the log-file read fully async and delete the synchronous object-store bridge. The other half of the original problem, making the base+log merge a `Stream`, is #685 and is not here - it owns the two remaining sync/async crossings and depends on this landing first.

### Batching the content fetch (slices 1 and 2)

Pass 3 issued one ranged read per admitted log block, serially, through the blocking bridge. By that point the gates have already chosen the set, so every range is known before any byte is read and they can go out together. `get_ranges` coalesces ranges that sit close together, so a run of adjacent blocks costs one round trip rather than one each. Nothing in the crate called `get_ranges` or `get_byte_ranges` before this.

The fetch moves into the async layer and Pass 3 keeps doing no I/O, because the record buffer's `process_data_block` is a synchronous trait method and cannot await.

Content is fetched and merged one window at a time, bounded by `hoodie.memory.dfs.buffer.max.size`, with a window's bytes dropped before the next is fetched. Fetching a whole slice's admitted content up front would give back the peak-memory property #682 was justified on (+200 MB to +40 MB on a 91 MB file): on a slice holding four 1 GB log files, every byte of it would be resident before the first record merged. A block larger than the budget still goes out on its own, since its content has to be whole to decode.

A window is a run of *consecutive* blocks sharing one file, because a `LogBlockFetcher` reads one file. Grouping by file first is enough for a bulk prefetch, whose result is keyed by index and consumed in the caller's order, but it cannot be interleaved with merging: Pass 3 merges oldest first, and with COMMIT_TIME_ORDERING a later block overwrites an earlier one for the same key, so a plan that reordered a slice's interleaved log files would let an older write win. Blocks with nothing to fetch keep their place, since the driver merges through each window's last index.

Measured against `main` with the fg-bench harness on a single-file-group MOR table (120k base rows, 130 MB of log content across 5 files, every base record updated), both refs built from the same source and run with identical flags in a 1-CPU container with swap disabled:

| memory cap | `main` | this PR |
|---|---|---|
| 384 MB | OK (324 MB peak) | OK (228 MB) |
| 320 MB | OK (326 MB) | OK (214 MB) |
| 288 MB | OOM-killed | OK (214 MB) |
| 256 MB | OOM-killed | OK (215 MB) |
| 224 MB | OOM-killed | OK (213 MB) |
| 192 MB | OOM-killed | OOM-killed |

The same read survives a container 30% smaller: the minimum viable cap moves from ~320 MB to ~224 MB. Both peaks are flat across caps, so each is the implementation's floor for this workload rather than budget-tracking, and each fails exactly where its floor exceeds the cap. On a log-heavy table (109 MB of logs) the unconstrained peak is 245 MB before and 150 MB after.

One window is in flight at a time, and within a window `get_ranges` merges only ranges less than 1 MiB apart and issues what remains ten at a time. So the coalescing helps where a file's admitted blocks are adjacent, and does nothing for blocks in separate log files. Windowing is latency-neutral against fetching everything up front: both await sequentially, so the total is the same sum of fetches and merges either way. What changes is only that the bytes are no longer all held at once.

### The header walk, and the bridge (slice 3)

The header walk was the last synchronous reader of object storage. It read through `std::io::Read` over `StorageReader`, whose window refills went through `get_range_blocking`: spawn onto a private runtime, wait on a `sync_channel(1)`. That bridge came with three non-local rules about where it could be called from - fine on a blocking-pool or plain thread, merely starving on another runtime's worker, a genuine deadlock on one of its own - none of which the compiler could check, and one of which needed a thread-name guard at runtime to refuse.

`StorageReader` now reads with async inherent methods. `read_bytes(len)` returns a `Bytes` slice of the whole-file buffer or of the resident window with no copy, `read_exact` fills a caller's array, and `seek_to` is pure cursor arithmetic that does no I/O at all - which is what most of the walk was doing through `Seek` anyway. A read longer than the window fetches exactly its own range and leaves the window alone, so a log block's content cannot grow what the reader keeps resident. With `Read` and `Seek` gone from the type, the `LogFileReader<R: Read + Seek>` generic had one instantiation and no reason to exist, so its 12 reading helpers became plain async methods over `StorageReader`.

Two decoding sites needed handling, because the decoders are synchronous and stay that way:

- The eager `read_all_blocks` used to hand the file cursor straight to `decode_content`. It now reads the block's own remaining span and decodes from a `Cursor` over it, then advances the reader by what the decoder actually consumed - which is not the block length, since a decoder is capped at the content length but need not read all of it. For the eager reader that span is a refcounted slice of the already-resident file, so it costs no bytes.
- Pass 3 called `LogBlock::load_content`, which fetched through the bridge. It now asks `queued_block_content` what to decode from: the prefetched bytes, nothing when the block already holds content, or an error. The prefetch admits exactly the block types Pass 3 decodes, so the error case means the block recorded nowhere to read from, which the old fetch would have failed on too. No I/O, so the pass stays synchronous and the scan future stays `Send`.

`get_range_blocking`, `OBJECT_STORE_RUNTIME`, its worker count and thread name, `in_object_store_runtime()` and the three call-site rules are all gone. Three comments elsewhere named the runtime and now describe the handle the blocking merge iterator is actually built with.

The known risk called out in #684 was the corrupt-block recovery scan, the one place the walk is not a forward parse: it scans forward for the next `MAGIC` over 1 MiB windows that overlap by `MAGIC.len() - 1`. It was prototyped first, and it came through unchanged in shape - the windows are now explicit `read_bytes` calls instead of `Read` + `Seek`, and the overlap arithmetic is the same.

- [ ] N/A
- [ ] Manual tests

**Batching.** Output is identical whether the prefetch works or not, so the batching needed tests that assert the mechanism rather than the result. Stubbing the prefetch to return nothing passed all 1309 existing tests. `test_read_contents_matches_reading_each_range_alone` asserts a batched read is byte-identical to reading the same ranges one at a time; `test_prefetch_returns_content_for_every_admitted_block` asserts every admitted block comes back with exactly its own range, and is the one that kills the stub-it-out mutation. The batching rules themselves had no test either, and removing the byte budget entirely passed all 1311 tests. The decision now lives in `plan_content_windows`, which returns the windows without doing any I/O, so five tests cover it directly: stopping at the budget, sending an oversized block alone rather than splitting or dropping it, never spanning files, skipping blocks with nothing to fetch, and emitting windows in consumption order. The last one fails against a plan grouped by file, which is what makes a window safe to merge before the next one is fetched.

**The corrupt-block scan.** `test_corrupt_recovery_scan_finds_a_magic_split_across_windows` lands the second block's `MAGIC` three bytes before the first scan window ends, and asserts the recovery offset itself rather than only the block count - for the eager reader and for fetch windows smaller than, larger than, and unrelated to the scan window. It was written and mutation-checked against the old synchronous code first, so it is a genuine before-and-after oracle rather than a test fitted to the new one.

**Block-skip arithmetic.** `test_skipping_a_block_lands_on_the_next_ones_magic` walks a two-block file with every block out of the instant window and asserts the cursor ends exactly at the file length. On a one-block file this cannot be tested: landing anywhere in the last few bytes reads as end-of-file either way.

Several of these need a log file with more than one block, and no shipped fixture has one - every file under `crates/core/tests/data/log_files` holds exactly one. They concatenate a one-block fixture with itself, which is a valid two-block file since log blocks are self-contained.

**Mutation checks**, each reverted afterwards: dropping the scan overlap, ignoring the end of a range when testing window coverage, removing the oversized-read path, not tracking the window start, reporting end-of-file as a different error kind, advancing past the block instead of by what the decoder read, disabling the batch budget, and swallowing a block with nothing to decode. Each failed a named test. One more - a one-byte error in the computed block end - passed the whole suite, which is what the two-block skip test above was then written for.

Verified locally: `make format`, the Rust half of `make check`, clippy on `hudi-core` with and without default features, the full `hudi-core` suite on both feature legs (1318 lib tests, 1298 with `--no-default-features`), the gold parity sweep green across all fixtures and both reader versions, and `cargo check` of `hudi-cpp` and `hudi-python`. This is a fork PR, so none of it ran in CI - the numbers above are local. The bounded-prefetch commit was re-verified on the `--no-default-features` leg (1299 lib tests), gold parity across both reader versions, clippy and `cargo fmt --check`; the default-features leg has not been re-run since it.

## What is not covered

Memory is measured (above); **wall time is not a claim**. Every local run has this PR 7-10% slower, which is a local-filesystem artefact: issuing all fetches up front lets OS readahead work the file sequentially, and interleaving fetch with merge breaks that. The effect should not exist against object storage, where there is no cross-request readahead to lose, but that is untested here.

The round-trip saving from batching is **not measured**, and it is regime-dependent: adjacent blocks within one log file coalesce, blocks spread across separate log files do not. `memory_bench` measures memory, not latency, and a local filesystem will not show the win at all, since it is a round-trip cost. Quoting a number needs a benchmark against real object storage, which is worth doing before this is claimed as a performance improvement rather than a structural one.

Overlapping a window's fetch with the previous window's merge is not done here and would be a separate change. Neither this PR nor its predecessor overlaps I/O with merge CPU at all, so the fetch-versus-merge share of wall time is the number that would say whether that second window of memory is worth spending.

The bindings still need runtimes of their own: Python's process-wide one and the cxx per-reader `current_thread` one do not go away because the core became async. What is gone is the runtime the *core* kept for its own blocking I/O.

## Downstream consumer (found in self-review)

`OBJECT_STORE_RUNTIME` was `pub`, and the `fg-bench` harness - a tracked crate at `benchmark/filegroup/` in `onehouseinc/hudi-rs-internal` - calls it, alongside `open()` and `open_blocking_stream()` which #688 removes. There is no caller in apache/hudi-rs, so nothing here breaks; the break lands when this stack merges down to the internal fork, where the harness stops compiling until it is ported to `open_stream`.

A port patch exists and is small (retarget two call sites; give the dedicated-thread mode its own `current_thread` runtime, which is what the cxx binding does per reader). Recommend landing it in hudi-rs-internal with the merge-down rather than after it.


yihua:
non-blocking: one question on object-storage performance before this stack's numbers get quoted. The memory win is measured and convincing, but every performance-shaped claim is local so far: the PR's own account says wall time is 7-10% slower on a local filesystem (attributed to lost OS readahead), the round-trip saving from `get_ranges` coalescing is unmeasured, and the header walk's windowed refills change the byte and round-trip profile in ways a local disk cannot show at all. Is an S3 (or other object-store) run of the fg-bench/memory-bench pair planned before the merge-down — ideally with a bytes-fetched counter, which would also answer the window-refill question on `fill_window` above in the same run?

yihua:

PR #688, opened 2026-08-19
refactor(core)!: merge base and log as a stream, deleting the sync bridge

linliu-code:
> **Stacked on #687 → #686 → #682 → #678. Review those first.** This targets `main`, so it shows their commits alongside its own. The delta that belongs to this PR is the last five commits: 8 files, +995/-765.

Second slice of #685, and the one that removes the crossings. With #686 (the log read) and this, a file-slice read crosses the sync/async boundary **zero** times instead of three.

## Description

The merge was a synchronous `Iterator`, so the async base file had to be turned back into a sync reader to feed it, and the sync result turned back into a `Stream` to return it. Both conversions are gone.

`FileGroupMergeStream` — renamed from `FileGroupMergeIterator`, because it is not one any more — produces a chunk with `next_chunk().await`, or becomes a `Stream` with `into_stream()`. Its base side is the parquet stream itself. The only await is the base pull; the merge kernel, the log-only drain and the output projection are synchronous work over data already in memory.

### What this deletes

| Deleted | Why it existed |
|---|---|
| `BlockingBatchReader` | Pulled the async base stream with `block_on` against a runtime handle; could only be driven off the worker threads. **Note for anyone searching:** #685 calls this `ParquetSyncReader`, a name that existed nowhere but in ten comments — also corrected here. |
| `open_blocking_stream` | Ran the whole merge on `spawn_blocking` behind a depth-1 channel, to bound the producer to one batch ahead of the consumer. A stream is demand-driven, so it has that bound by construction. |
| The `streaming: bool` split | `make_base_file_source` returned the same type from both branches with two different safety contracts — one "MUST be consumed in sync context", the other safe from async callers. There is one base source shape now, and it is always async. |
| `HoodieFileGroupReader::open()` | No caller **in apache/hudi-rs** - not core, hudi, datafusion, python or cpp - and the cxx bridge reads through `read_file_slice*` with its own runtime. It is **not** callerless everywhere: `fg-bench`, a tracked crate in `onehouseinc/hudi-rs-internal`, calls it and `open_blocking_stream()`, reaching them through an uncommitted `pub mod reader_v2` exposure kept in that fork. See *Downstream impact* below. |
| `apply_instant_range_filter` | It never inspected a batch — it kept all of them or none, by the base file's own commit instant. The decision moved ahead of the read as `base_file_in_range`, so an excluded base file is no longer read and discarded, and the eager fallback that existed only to materialise rows for that filter is gone with it. |

### `read()` no longer collapses the base (changed after self-review)

An earlier revision of this PR kept `read()` merging the base as a single batch, via a `collapse_base` step, to preserve its output row order: the merge kernel emits surviving base rows followed by the log records that replaced the rest, so splitting the base across row groups interleaves the two per group. Same rows, different sequence.

Self-review asked what pinned that order, and the answer was nothing - the gold sweep sorts before comparing, and the eager-vs-stream test compared only counts. Measuring instead of arguing reversed the decision twice over:

- **The order argument does not reach any table we have.** It needs a base batch holding both surviving and replaced rows; every MOR fixture in this repo and both benchmark datasets update every key, so no base row survives to be ordered against. Reading `churn-smoke` and `logheavy-smoke` both ways gives byte-identical sequences (120k and 60k rows, 118 and 59 streamed chunks against one collected batch).
- **Collapsing costs memory**, because it materialises the whole base file before merging. Peak RSS on `churn-smoke`: **304 MB with it, 258 MB without** - 15% on a 142 MB dataset.

So `collapse_base` is gone. Both entry points merge a row group at a time and return the same sequence by construction - one behaviour instead of two - and `read()` loses a full copy of the base file from its peak.

### Downstream impact

`open()` and `open_blocking_stream()` are called by `fg-bench` in `onehouseinc/hudi-rs-internal` (tracked, at `benchmark/filegroup/`), together with `OBJECT_STORE_RUNTIME` which #686 deletes. Nothing in apache/hudi-rs breaks; the harness stops compiling when this stack merges down to that fork. A small port patch exists (retarget the call sites to `open_stream`; give the dedicated-thread mode its own `current_thread` runtime, as the cxx binding does per reader) and should land with the merge-down.

### A silent gap in the stats, documented not closed

`read_stats()` is complete after `read()` and incomplete after `open_stream()`: the merge-phase counters (timings, peak map entries, insert/update/delete counts) accumulate in the shared handle as the stream is consumed, and nothing folds them back, because the caller owns the stream and the reader never learns when it ended.

It predates this PR and is not closed here - no production caller reads these, only the test harness and the benchmark do. It is documented on `read_stats()` because the gap reads as data rather than as absence: a streaming read of a fixture with five deletes reports `num_deletes: 0` while returning exactly the rows the eager read returns, and a zero in a benchmark gets believed.

### Deliberately deferred

The buffer trait now expresses "merge a base batch" twice. `next_merged_base_batch`, `pull_and_merge_next_base_batch`, the `base_file_source` field and `set_base_file_source` have **zero production callers** after this change (three call sites, all tests), and `has_next` / `next` / `merge_and_collect_with_stats` were already production-dead - the streaming path carries `unreachable!` for the last of them. Removing that second API is a mechanical follow-up, kept out of this diff to keep the deletion reviewable; recorded here so it is a decision rather than an oversight.

### One stat corrected

`final_merge_ms` now times only the merge. Timing the loop, which is what the code shape invites, would fold the base file's read latency into a stat that means merging.

- [ ] N/A
- [ ] Manual tests

Two coverage gaps surfaced, both the same shape — **fewer rows, no error**:

- **`test_every_base_row_group_reaches_the_output`** reads a base file written in four row groups, through both entry points. Making `collapse_base` keep only the first row group **passed all 1321 tests** before this test existed: every base file in the suite fits in one row group, so keeping one and keeping all look identical. The stream side asserts more than one chunk, which is what proves the fixture really spans several groups.
- **`streaming_and_eager_reads_agree`** compared row counts only, which passes for a stream returning the right number of wrong rows. It now compares row content, sorted — sorted because the two entry points chunk the base differently and Hudi promises no row order.

The schema-evolution test no longer has an eager and a streaming source to compare, so it compares the base source against its collapsed form instead — still two shapes, still required to agree.

Two more from the self-review, pinning the order property that `collapse_base` used to defend implicitly:

- **`batch_boundaries_change_the_merged_row_order`** merges a base holding both surviving and replaced rows, whole and then split in two, asserting `a c e b d` against `a c b e d` - the same rows in a different order. Its first version asserted nothing, because `extract_records` sorts by key; it now reads the key column in row order.
- **`test_stream_merges_a_slice_incrementally_and_matches_the_eager_read`** compared *sorted* keys, so it passed whatever the order was. It now compares the sequence. Stated in the test rather than implied: it cannot detect a violation on today's fixture - reintroducing a collapse on `read()` alone leaves it green, confirmed by mutation - because that fixture updates every key. It pins the two paths together and will discriminate given a fixture with surviving base rows across more than one batch.

### Is it actually a stream, though?

Row and order assertions cannot answer that: a merge that collected the base up front, or one that ran on a blocking-pool thread feeding batches back through a channel - the shape this PR replaces - returns the same rows in the same order. So `merge_stream_is_lazy_and_polls_on_the_callers_thread` asserts the two things that differ:

- **Lazy** - one chunk costs one base pull, counted, not the whole base file.
- **No thread hand-off** - every base pull lands on the thread polling the stream, checked by thread id on a `current_thread` runtime.

- **Non-blocking** - a concurrent ticker task advances *while each chunk is produced*, on a `current_thread` runtime where the only way that can happen is the merge giving the thread up (`merge_stream_lets_other_tasks_run_between_chunks`).

All three mutation-checked, each reverted:

| Mutation | Outcome |
|---|---|
| Pre-collect the base | first chunk costs 3 pulls, not 1 |
| Pull through `spawn_blocking` | pulls land on a different thread id |
| Await the base up front, then serve chunks without awaiting | ticker stalls at chunk 2 (`4 -> 4`) |
| Block the thread *through* the base await | **deadlocks** - the test hangs rather than failing |

The last one is worth stating plainly: the most severe violation is detected only as a hang, because on one worker a pending read can never be rescheduled. Detected, but a poor signal, so the per-chunk interleaving check is what that test rests on. Recorded in the test's own comment too.

Worth noting what does *not* prove any of this - all 54 `read_stream` tests run on a single-worker `current_thread` runtime, but so did the `spawn_blocking` design, so that alone was never evidence.

Mutation-checked, each reverted afterwards:

| Mutation | Result |
|---|---|
| `collapse_base` drops all but the first row group | fails only `test_every_base_row_group_reaches_the_output` |
| No-merge path forwards only the first base batch | fails 4 tests |
| Invert the hoisted instant-range decision | fails 8 tests |

Verified locally: format, clippy on `hudi-core` with and without default features and over all targets, the full `hudi-core` suite both ways (1322 lib tests, 1302 without defaults), the gold parity sweep green across all fixtures and both reader versions, and `cargo check` of `hudi-cpp` and `hudi-python`. This is a fork PR, so none of it ran in CI — the numbers above are local.

## Known, deliberate, and next

A merge map that has **spilled** probes RocksDB synchronously, and that probe now runs on an async worker rather than a blocking-pool thread. It is local disk, and only past `hoodie.memory.merge.max.size`. This is the caveat #685 raises itself, and the honest position is that it needs a measurement before a fix: gating it on `merge_map_spilled()` (already on the buffer trait) is cheap, but adding `spawn_blocking` per batch on principle would reintroduce a crossing this PR exists to remove.

The bindings still keep runtimes of their own — Python's process-wide one, the cxx per-reader `current_thread` one. Those are #685's stated non-goals and do not go away because the core became async. What is gone is the runtime the *core* kept for its own blocking I/O (#686) and the blocking thread it kept for its own merge (here).

Also worth recording, not fixed: `hoodie.read.stream.batch_size` is honoured for a base-file-only slice and has no effect on a slice with log files, where chunking follows the base file's row-group cadence. Flagged in #687; unchanged here.


yihua:

PR #689, opened 2026-08-19
perf(core): bound the merge chunk rather than moving the merge off the executor

linliu-code:
> **Stacked on #688 → #687 → #686 → #682 → #678. Review those first.** This targets `main`, so it shows their commits alongside its own. The delta that belongs to this PR is the last commit: 3 files, +197/-4.

Third and last slice of #685, and it closes the one question the ticket left open. **The answer turned out to be "don't add `spawn_blocking`"** — so this PR is a measurement, a decision, and the bound that makes the decision hold.

## The question

#685 flagged it itself: a merge map that has spilled probes RocksDB synchronously, and once the merge is a `Stream` (#688) that probe runs on the task polling it rather than on a blocking thread. The ticket's own instruction was to measure before adding `spawn_blocking` for its own sake, "since that would reintroduce a crossing".

## The measurement

`spilled_merge_blocking_duration` (ignored; run with `--release --ignored --nocapture`) times one `merge_base_batch` against a 50 000-key log map, with and without the spill engaged:

| chunk rows | no-spill | spilled |
|---|---|---|
| 1024 | 0.4 – 1.1 ms | 5.7 – 6.1 ms |
| 8192 | 2.8 ms | 39 – 42 ms |

**The no-spill leg is the control, and it is what decides the question.** My first version of this benchmark did not have it, and without it the 40 ms figure looks like a spill problem. With it, the spill is a **~6× amplifier on work that is already synchronous and of the same kind**. Gating `spawn_blocking` on `merge_map_spilled()` (already available on the buffer trait) would treat the amplifier and leave a poll that is milliseconds long either way.

Per-chunk CPU work on the polling task is how a scan or join operator behaves. Handing each chunk to a blocking pool would reintroduce the sync/async crossing #686 and #688 removed, and cost a task hop plus a buffer move per chunk, to shave single-digit milliseconds.

## What the numbers do say

The cost is linear in a chunk's rows. That makes the chunk size the thing bounding how long one poll occupies its executor — and that bound was **inherited, not chosen**: the base read set no batch size, so it came from `parquet`'s own default of 1024 (`parquet-57.3.1`, `arrow_reader/mod.rs:171`, verified rather than assumed). That is a silent bound. A larger default upstream, or a caller passing a bigger batch size through here, multiplies every poll's cost with nothing failing.

`base_read_options` now asks for it explicitly as `MERGE_CHUNK_ROWS`.

**Behaviour change, small but real:** a parquet base file is unaffected — 1024 is what it was already getting. A **lance** base file defaulted to 8192, so its merge chunks become 1024: the same rows in more chunks, and the bound is now uniform across base-file formats instead of differing eightfold by backend. `read_data` concatenates its stream, so the single-batch path returns exactly what it did before.

- [ ] N/A
- [ ] Manual tests

`test_a_merged_chunk_is_bounded_by_the_readers_own_batch_size` writes 5000 rows in **one** row group and asserts no chunk exceeds the bound — so a chunk that followed the file's layout fails it. It also asserts the option directly, and that combination is deliberate: the bound agrees with parquet's default today, so no output-level assertion can distinguish the pin from the default, while the realistic future break is a caller passing a larger size (for instance making `hoodie.read.stream.batch_size` effective on the merge path — the gap recorded in #687).

Mutation-checked, each reverted: dropping the pin, and raising it eightfold. Each fails that test.

Verified locally: format, clippy over all targets and without default features, the full `hudi-core` suite both ways (1324 lib tests, 1303 without defaults), the gold parity sweep green across all fixtures and both reader versions, and `cargo check` of `hudi-cpp` and `hudi-python`. Fork PR, so none of it ran in CI — the numbers are local.

## Limits of the measurement

Stated so nobody reads more into it than it supports:

- The spill file is freshly written, so the page cache is warm and **the disk component is understated**. A cold spill file would be slower — though still bounded by the same chunk size, which is the point of pinning it.
- One machine, single-threaded, release build. It is enough to separate "milliseconds" from "hundreds of milliseconds", which is the distinction the decision rests on, and not enough to quote as a throughput figure.

If the spilled path ever does need to come off the executor, the hook is unchanged and cheap: `merge_map_spilled()` on the buffer trait, checked once per merge rather than per chunk.

## A documented knob that does nothing on this path (found in self-review)

`docs/reader-spec.md` documents `with_batch_size(n)` -> `hoodie.read.stream.batch_size`, default 1024, and Python exposes it (`ReadOptions::with_batch_size`, `options.batch_size()`). It is honoured for a base-file-only slice and has **no effect** on a slice with log files, where the chunk follows `MERGE_CHUNK_ROWS`.

Measured, not argued: `--batch-size` 256 / 1024 / 8192 over a 1.43M-row table gives 1174 / 1164 / 1169 ms and identical rows. Flat.

This predates the stack - the value was already read and discarded - but removing the last reader of it and pinning the effective bound at a constant makes the divergence permanent, so it belongs on the record here.

**`docs/reader-spec.md` is corrected in this PR**: the row now reads `1024` (streaming, base-file-only slices), with a note that a merged slice sizes its own chunks and this knob has no effect there. The behaviour is left alone deliberately - making the knob effective on the merge path would scale how long one chunk of merging occupies its executor, linearly per the numbers above, so it is a decision of its own rather than a doc fix.

### DIFF
--- a/crates/core/src/file_group/reader_v2/engine.rs
+++ b/crates/core/src/file_group/reader_v2/engine.rs
@@ -416,13 +416,17 @@ impl HoodieFileGroupReader {
     /// Stream the merged output.
     ///
     /// [`Self::read`] returns the whole file group as one batch, so peak memory
-    /// tracks the base file. This reads the base file one row group at a time
-    /// instead, merging and emitting a chunk per group.
+    /// tracks the base file. This reads the base file one bounded batch at a
+    /// time instead (`MERGE_CHUNK_ROWS` rows), merging and emitting a chunk
+    /// per batch.
     ///
     /// Demand-driven: nothing is merged until the consumer asks for it, so the
     /// memory this adds over the merge map is one chunk. The previous shape ran
     /// the merge on a blocking thread behind a depth-1 channel to get the same
     /// bound; a `Stream` has it by construction.
+    ///
+    /// Single-use: takes the output converter and, for MOR, moves the record
+    /// buffer into the returned stream. Reading again needs a new reader.
     pub(crate) async fn open_stream(
         &mut self,
     ) -> Result<futures::stream::BoxStream<'static, Result<RecordBatch>>> {
@@ -436,11 +440,13 @@ impl HoodieFileGroupReader {
     /// `RecordBatch`.
     ///
     /// Same merge as [`Self::open_stream`], collected into one batch. Both
-    /// entry points merge the base a row group at a time and therefore return
+    /// entry points merge the base a batch at a time and therefore return
     /// the same row sequence; this one just concatenates the chunks.
+    /// Single-use, like [`Self::open_stream`].
     pub async fn read(&mut self) -> Result<RecordBatch> {
-        // Stage timing (perf harness): the base file open plus, here, its whole
-        // decode — `collapse_base` drives the stream to completion.
+        // Stage timing (perf harness): only the open, same as `open_stream` —
+        // the decode happens lazily while `collect_into_one_batch` drives the
+        // stream, so it lands in the merge loop rather than in `base_read_ms`.
         let base = profile_once!(self.read_stats.base_read_ms, self.base_file_source().await)?;
         let batch = self
             .init_record_iterators(base)
@@ -684,10 +690,10 @@ impl HoodieFileGroupReader {
     /// Open the base file as a stream of batches, with the schema they carry.
     ///
     /// One shape for every caller: the base file is read asynchronously, one
-    /// row group at a time, and the whole file is never resident. A caller that
-    /// needs it as a single batch collapses it afterwards
-    /// ([`Self::collapse_base`]) — that is a choice about chunking, not about
-    /// what is safe to call from where.
+    /// bounded batch at a time, and the whole file is never resident. A caller
+    /// that needs it as a single batch collapses it afterwards (`read()` does,
+    /// via `collect_into_one_batch`) — that is a choice about chunking, not
+    /// about what is safe to call from where.
     ///
     /// Returns an empty stream when the input split has no base file (log-only
     /// file group), and when the instant range excludes this base file: the
@@ -834,7 +840,7 @@ impl HoodieFileGroupReader {
         }
 
         // Open the base file as a stream. The whole file never lives in memory;
-        // one row group does. The (CoW-gated) RowFilter is threaded through the
+        // one batch does. The (CoW-gated) RowFilter is threaded through the
         // intersection read so row groups can be pruned via column-index stats
         // (the builder resolves predicate columns by name and returns None when
         // any referenced column is absent — safe even for evolved/added cols).
--- a/crates/core/src/file_group/reader_v2/engine.rs
+++ b/crates/core/src/file_group/reader_v2/engine.rs
@@ -37,7 +37,7 @@ use crate::file_group::reader_v2::buffered_record_converter::BufferedRecordConve
 use crate::file_group::reader_v2::input_split::InputSplit;
 use crate::file_group::reader_v2::iterator_mode::IteratorMode;
 use crate::file_group::reader_v2::merge_iterator::{
-    FileGroupMergeIterator, StreamStatsHandle, new_stream_stats_handle,
+    FileGroupMergeStream, StreamStatsHandle, new_stream_stats_handle,
 };
 use crate::file_group::reader_v2::output_converter::OutputConverter;
 use crate::file_group::reader_v2::profiling::profile_once;
@@ -48,6 +48,7 @@ use crate::file_group::reader_v2::schema_handler::FileGroupReaderSchemaHandler;
 use crate::storage::{RowFilterBuilder, Storage};
 use arrow_array::RecordBatch;
 use arrow_schema::SchemaRef;
+use futures::{StreamExt, TryStreamExt};
 use std::str::FromStr;
 use std::sync::Arc;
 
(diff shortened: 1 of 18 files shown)


crates/core/src/file_group/reader_v2/engine.rs
```diff
@@ -133,9 +134,25 @@ pub struct HoodieFileGroupReader {
...
+/// 0.4-1.1 ms, and 5.7-6.1 ms once the merge map has spilled to disk; at 8192
+/// rows those become 2.8 ms and ~40 ms. So the chunk size is what bounds how long
+/// a single poll occupies its executor, and it is set here rather than inherited.
+///
+/// 1024 is what `parquet` already defaults to, so this pins today's behaviour
+/// instead of changing it. Pinned because the bound is silent if it moves: a
+/// larger default upstream would multiply the blocking above with nothing
+/// failing. Measured by `spilled_merge_blocking_duration` (ignored; run with
+/// `--release --ignored --nocapture`).
+const MERGE_CHUNK_ROWS: usize = 1024;
```
yihua, line 154: non-blocking: The chunk bound pinned here is 1024 for merged base chunks, but the log-only drain still emits 4096-row chunks per poll (`DEFAULT_BATCH_SIZE` in `drain_log_only_inserts`), and a spilled map's drain does its RocksDB pulls on the same polling task the bound exists to protect. Was leaving the drain at 4x the merge bound deliberate — e.g. because the drain has no per-row map probe — or should it follow `MERGE_CHUNK_ROWS` too? Either answer seems fine, but the asymmetry is the one occupancy decision the PR doesn't record.

crates/core/src/storage/reader.rs
```diff
@@ -240,62 +223,113 @@ impl StorageReader {
...
+        self.pos = pos;
+    }
+
+    /// Whether `[start, end)` lies inside the resident window.
+    fn window_covers(&self, start: u64, end: u64) -> bool {
+        start >= self.window_start && end <= self.window_start + self.window.len() as u64
+    }
+
+    /// Fetch a window starting at the cursor.
+    async fn fill_window(&mut self) -> Result<()> {
```
yihua, line 249: non-blocking: Two fetch patterns here look worth a measurement before the merge-down: on a file whose blocks are smaller than the window, the walk's refills sweep essentially the whole file and Pass 3 then re-fetches the admitted content, so total bytes read are ~2x the old whole-file GET; and for a block larger than the window, `is_block_corrupted`'s trailing-pointer probe evicts and re-fetches the header-side window (seek to trailing → refill → seek back → refill again), costing ~2 extra windows per large block. Have you considered a bytes-fetched counter in the object-storage bench you already plan, or restoring/keeping the walk window across the corruption probe? The memory win is measured and real — this is only about knowing what it costs in bandwidth where round trips were the justification.

crates/core/src/file_group/reader_v2/merge_iterator.rs
```diff
@@ -487,22 +554,23 @@ impl Iterator for FileGroupMergeIterator {
...
-}
 
-impl RecordBatchReader for FileGroupMergeIterator {
-    fn schema(&self) -> SchemaRef {
-        self.output_schema.clone()
+    /// Drive this merge as a `Stream`, for a caller that wants to compose it.
+    ///
+    /// Chunk-at-a-time and demand-driven: nothing is merged until the consumer
+    /// asks, so the extra memory over the merge map itself is one chunk.
+    pub fn into_stream(self) -> BoxStream<'static, Result<RecordBatch>> {
```
yihua, line 566: non-blocking: `unfold` panics if polled again after it returns `Ready(None)`, and this stream is the public surface of `read_stream`. `next_chunk`'s sticky `done` already makes the extra poll semantically a no-op, so a `.fuse()` before `.boxed()` would turn that panic into the `None` the caller expects, for free.

crates/core/src/file_group/reader_v2/merge_iterator.rs
```diff
@@ -376,92 +397,138 @@ impl Iterator for FileGroupMergeIterator {
...
+                                    Some(Ok(b)) => b,
+                                    Some(Err(e)) => {
+                                        // Drop the source: a failed read must not
+                                        // be resumed as if it had merely ended,
+                                        // which would truncate the output
+                                        // silently.
+                                        *base_source = None;
+                                        log::error!("[FileGroupMergeStream] base file source: {e}");
+                                        chunk_err = Some(CoreError::ReadFileSliceError(format!(
+                                            "base file source error: {e}"
```
yihua, line 468: non-blocking: The Buffered path stringifies a base-source error into `ReadFileSliceError` while the Eager path forwards the original `CoreError`, so the same storage failure surfaces as a different variant depending on whether the slice had log files. Forwarding `e` unchanged here (and in `pull_and_merge_next_base_batch`) would keep the taxonomy uniform and preserve the source chain for callers that classify retryable errors.

yihua:
non-blocking: one question on object-storage performance before this stack's numbers get quoted. The memory win is measured and convincing, but every performance-shaped claim is local so far: the PR's own account says wall time is 7-10% slower on a local filesystem (attributed to lost OS readahead), the round-trip saving from `get_ranges` coalescing is unmeasured, and the header walk's windowed refills change the byte and round-trip profile in ways a local disk cannot show at all. Is the performance difference lower and could there be benefit on S3 (or other object-store) given that object storage has higher chunk read latency?

yihua (approved):
Overall LGTM

PR #691, opened 2026-08-25
feat(core): read HFile base files through ranged, window-coalesced reads

linliu-code:
## Description

**Stacked on #689, review that one first.** Until it merges, the diff here shows the stack's commits as well as this one; the delta is `a744c12..HEAD`, six files. Replaces #690, which implemented the same capability by holding the whole file.

The version two file group reader selects a base file reader by format, but `create_base_file_reader` refuses HFile, so a metadata table file slice cannot be read through it. This adds an `HFileBaseFileReader` and selects it inside the version two reader. The shared factory arm is left refusing, so the legacy reader and the two tests that pin that refusal are unaffected.

The decoder now reads through a byte source rather than one buffer. A resident source slices bytes already in hand, which is what an HFile arriving as a log block's content is; a ranged source reads the trailer and the load-on-open section up front, then fetches data blocks as they are read. Data blocks are grouped into runs that stay under `hoodie.memory.dfs.buffer.max.size` and read one request per run, so a full scan costs a request per window rather than one per block, and peak memory tracks one window rather than the file. The multi-level index walk fetches a level per request for the same reason.

A third commit decodes each HFile value rather than handing it over as bytes, because a base file and a log block of the same table must yield the same columns to merge. Two schema details needed care: the Avro to Arrow converter has no support for named type references, which the metadata table's record schema uses, so the Arrow schema is taken from the decoder; and the parsed schema's canonical form rewrites those references into something the decoder cannot resolve, so the writer's original JSON is read instead of round tripped. A record's key field is filled from the HFile entry key when the writer left it empty, mirroring `HoodieAvroHFileReaderImplBase`, including its two guards: a populated value and a null value are both left alone. That field name is a coupling to the metadata table's schema, documented as such in `crates/core/src/hfile/record_key.rs`.

A fourth commit stops reading every HFile in ranges. A ranged open pays a metadata lookup, a trailer range and a load-on-open range before the first data byte, so on a metadata base file of a few kilobytes it is four requests where one would do. Hudi reads the file whole below `hoodie.metadata.file.cache.max.size.mb` and opens a seekable stream above it, and this now does the same with the same key and the same 50 MB default. The size comes from the caller when the listing already supplied it, and otherwise from the reader the ranged path would open anyway, so the decision costs no request of its own.

Scope is the base file only. Merging HFile log blocks and key or prefix pushdown are each separate changes. `MetadataTableFileGroupReader` and the log block decode path still read resident bytes and are unchanged here.

No latency measurement is claimed. The window sizing follows from request-count arithmetic and from the pattern the log file path already uses, not from a benchmark; proving it on real data needs a production metadata table HFile with a multi-level index, which no fixture here provides.

- [ ] N/A
- [ ] Manual tests

Three additions. Every fixture under `crates/core/tests/data/hfile/` is read both ways and the records compared, which is the guard on the two multi-level index walks being separate code paths; the deep index fixture is the one that exercises the split. A metadata table `files` partition base file is read through the version two reader and its key set asserted equal to what `MetadataTableFileGroupReader` returns for the same base file only slice, failing before this change with `UnsupportedBaseFileFormat`. Window planning is covered directly, including that a block larger than the budget is still read whole.

The read strategy is asserted by request count rather than by rows, because both strategies return the same rows over the same file: a store wrapping the real one counts what passes through, and the same file reads as one request with the size known, one request plus one lookup without it, and three requests plus a lookup with the threshold at zero. Breaking the decision so it always reads in ranges fails that test with `left: (3, 1)  right: (1, 0)`.

A second commit adds a case built from a fixture's trailer alone, so its load-on-open offset points past the end of the file. Unguarded that underflows the length passed to the ranged read, which panics in debug; the test fails with `attempt to subtract with overflow` without the guard.

Run locally: 1337 pass with default features, 1317 with `--no-default-features`, clippy clean on both feature sets with warnings denied, fmt clean, and `cargo check --workspace --all-targets --all-features` clean including the Python and C++ bindings. No CI run has happened on this branch: fork pull requests sit at `action_required` until a committer approves them, so all of the above is local only.


yihua:

PR #692, opened 2026-08-25
test(core): cover every metadata partition's log blocks through reader v2

linliu-code:
## Description

**Stacked on #691, which is stacked on #689. Review those first.** Until they merge the diff here shows their commits too; the delta is `50ae752..HEAD`, three files.

A file slice holding an HFile log block reads as though the slice had no log blocks: the read succeeds and returns the base file's rows. Nothing errors, and no counter records a skipped block, so the loss is silent. Version two is the default reader, and the version one fallback does not cover this because version two never reports an inability, it reports success.

The block type was skipped at five points: the prefetch window planner, pass one's classification into the per-instant map, pass three's decode dispatch, pass three's buffer dispatch, and the buffer itself, whose `if let LogBlockContent::Records(..)` has no else and discards anything other shape. Admitting the type at the first four alone would not have helped, because the buffer takes Arrow and nothing else.

An HFile block's records are Avro values stored under HFile keys, so they are decoded against the schema in the block header, which is the same resolution an Avro block already gets, and the buffer receives batches like every other data block type. That resolution now lives in one helper shared by both paths so they cannot drift on promotions. Whether an HFile block is decoded or handed back as key-value pairs is the caller's choice, because the metadata table reader wants the pairs.

A writer may leave the record's key field empty because the HFile entry key already holds it, so an empty field is filled from the entry key. This shares one implementation with the base file reader added in #691, including its guards: a populated value and a null value are both left alone.

- [ ] N/A
- [ ] Manual tests

`harness_hfile_log_block_rejected` asserted `Invalid block type: 4`, which held upstream where HFile blocks were unrecognised, and was ignored with a note that the expectation needed re-deriving rather than inverting. It is re-derived and un-ignored as `harness_hfile_log_block_merges`.

Its expectation comes from the inputs, not from this reader: the base file was read with the parquet reader, and the block's two records were decoded with `apache_avro::from_avro_datum` against the schema in the block header. Both sides carry `ts = 100`, so event-time ordering ties and the case's commit-time ordering decides, which makes the log write win on two of the four keys. Without the fix the read returns the base file unchanged, so the case fails on the commit time of the first row.

The `key` column is asserted because it is where the entry key is filled. The base file holds it and the log writer left it empty, so without the fill two rows read as empty next to two holding their own keys.

This is **not** a Hudi parity oracle. The fixture has no gold data, because Spark cannot read it either, so nothing here checks hudi-rs against Hudi; it checks the merge against the decoded inputs.

The metadata table is the reason this change exists, so it is read directly rather than left to the next change in the stack, and every partition in the fixture is read rather than the `files` partition alone: a skipped block is silent, so a `files`-only test would pass on four of the five partitions regardless. `column_stats`, `partition_stats` and `record_index` compact to a base file written after all their log files, so their pre-compaction slice has log files and no base file at all, which makes them the sharper case: a skipped block leaves those reads with no records and no schema to build an output from, and they fail outright. `files` and `secondary_index` have both, so there the merged read is asserted to differ from a base-file-only read. The `files` slice additionally has a real oracle, `MetadataTableFileGroupReader`, which reads the same slice through a separate implementation on decoded structs rather than Arrow.

The merge mode used is commit-time ordering, which is not the metadata table's own mode. These cases pin that the blocks are decoded and reach the merge; whether the merge applies the metadata payload's rule is a separate change.

The test's metadata configs gained the table's own record key config. Without it the key resolves to `_hoodie_record_key`, which every metadata record leaves empty, so a merge collapses every key onto one entry. A base-file-only read never noticed, because with no log files there is no merge, which is why the gap survived until a test read log files.

Run locally: 1339 pass with default features, 1319 with `--no-default-features`, clippy clean on both feature sets with warnings denied, fmt clean, and the metadata table reader's own test still passes, which is what confirms that path still receives raw key-value pairs. Each of the four sites this change admits the block type at was reverted in turn, and the metadata cases fail on each. No CI run has happened: fork pull requests sit at `action_required` until a committer approves them, so all of the above is local only.

### DIFF
--- a/crates/core/src/file_group/base_file/hfile.rs
+++ b/crates/core/src/file_group/base_file/hfile.rs
@@ -1584,4 +1584,230 @@ mod tests {
         );
         Ok(())
     }
+
+    /// One metadata-table file slice: the partition it lives in, its base file if
+    /// the slice has one, and the log files written after it.
+    struct MdtSlice {
+        label: &'static str,
+        partition: &'static str,
+        base_file: Option<&'static str>,
+        log_files: &'static [&'static str],
+        /// The base instant the slice is read as of.
+        instant: &'static str,
+    }
+
+    /// Every metadata partition in the fixture that has a merge-reachable slice.
+    ///
+    /// `column_stats`, `partition_stats` and `record_index` compact to a base file
+    /// written *after* all their log files, so their pre-compaction slice has log
+    /// files and no base file at all. That is the sharper case for this change: an
+    /// HFile log block that is skipped contributes nothing, so the read has no
+    /// records and no schema to build an output from, and fails outright.
+    const MDT_SLICES: &[MdtSlice] = &[
+        MdtSlice {
+            label: "files",
+            partition: "files",
+            base_file: Some("files-0000-0_0-955-2690_00000000000000000.hfile"),
+            log_files: &[
+                ".files-0000-0_20251220210108078.log.1_10-999-2838",
+                ".files-0000-0_20251220210123755.log.1_3-1032-2950",
+                ".files-0000-0_20251220210125441.log.1_5-1057-3024",
+                ".files-0000-0_20251220210127080.log.1_3-1082-3100",
+                ".files-0000-0_20251220210128625.log.1_5-1107-3174",
+                ".files-0000-0_20251220210129235.log.1_3-1118-3220",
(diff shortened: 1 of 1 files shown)


yihua (commented):
LGTM.  Thanks for enhancing the MDT test coverage!

yihua (approved):

PR #696, opened 2026-08-27
fix(core): decode a log block that holds more than one batch of records

linliu-code:
## Description

**Stacked on #695, which is stacked on #694, #693, #692 and #691. Review those first.** Until they merge the diff here shows their commits too; the delta is `eede8d9..HEAD`, ten files.

A log block whose record count exceeds the decoder's batch size fails to decode at all: `Offsets must be non-negative and within the length of the Array`. The batch size is 1024 and a metadata table's log blocks exceed that routinely, so the version two metadata read cannot handle a realistic metadata table. Nothing in the test corpus saw it, because every metadata fixture block holds two or three records and the largest data fixture of any kind is 458 KB.

The trigger is any Avro union with two or more non-null branches, which Arrow represents as a dense union, and which the metadata record schema always carries in its column statistics. Unions spelled `["null", T]` are unaffected: Arrow makes those a nullable `T`, with no offsets buffer. `arrow-avro`'s union decoder drains its offsets on flush but keeps the per-branch counts that produced them, so the second batch emits offsets past children it has just emptied. Reported upstream as apache/arrow-rs#10876 with a three-line reproducer. Its decoder exposes no reset, so a clean one has to be built, which now happens after a mid-block flush. A block that fits inside one batch never reaches that path and pays nothing.

Rebuilding per batch would otherwise repeat schema-sized work, so the writer schema's parse and fingerprint are cached and reused across blocks. Measured on the metadata record schema, which is 7,969 bytes of JSON, that registration is 234us of the roughly 400us a construction costs. What remains is the two parses inside `arrow-avro`'s own decoder construction, which its public API gives no way to skip.

Separately, the header walk now keeps a block's content when its fetch window already covered it, instead of seeking past and re-requesting the same bytes in pass three. That is one request per log file rather than two, matching what the eager reader has always cost. The bytes are copied rather than sliced, and charged against a budget of one window shared across the whole scan: a `Bytes` slice pins its entire window allocation, and the walk accumulates blocks from every log file before the gates run, so residency would otherwise grow with the slice instead of staying at the bound the headers-only walk exists to buy.

Stage timings move from milliseconds to microseconds. A metadata read completes in about a millisecond, so every stage rounded to zero and the counters said nothing. Four timers that wrapped an empty span are removed: the fetch they were written for had moved to pass three, leaving `log_block_decode_ms` structurally zero. The log path is now split into fetch, decode and upsert, which is what located the cost described above.

- [ ] N/A
- [ ] Manual tests

Three additions and two re-derivations.

The decode fix has two tests over a three-line hand-written schema with a two-branch union, encoding 100 records at a batch size of 16, so they need no fixture and no generated data. One asserts every row returns and that more than one batch was actually produced, so it cannot pass vacuously if a batch-size change ever makes 100 records fit in one. The other alternates the union's branches and checks every row's identity across the boundary, because a corrupt offsets buffer returns wrong values rather than only crashing. A companion test without a union pins that multi-batch decoding is not broken in general. With the rebuild removed, both union tests fail with the upstream error and the non-union test still passes.

The one-request-per-log-file change is asserted by request count through a store that wraps the real one, not by rows returned: both paths return the same records over the same file, which is the point. Inside one window every block is resident and the file costs one get; with a 64 byte window every block defers and it costs more than one. A separate test pins that the residency budget is a bound: a generous budget lets the walk keep content, a zero budget keeps none, and a budget below the total keeps less than everything.

Two existing tests changed subject and were re-derived rather than inverted. `test_load_content_reports_a_block_running_past_the_file_end` truncated its file after the walk, which no longer leaves a ranged read to come back short; it now forces deferral with a one byte window and keeps its assertion, plus a guard that deferral actually happened. `test_prefetch_returns_content_for_every_admitted_block` had become vacuous, since both of its blocks were resident and `fetch_window` was never called; it now walks with a zero residency budget and asserts nothing is resident, so the keying it exists to pin is exercised again.

Run locally: 1385 pass with default features, 1365 with `--no-default-features`, clippy clean on both feature sets with warnings denied, fmt clean, and `cargo check --workspace --all-targets --all-features` clean including the Python and C++ bindings. That last one matters here: `HoodieReadStats` is public and seven of its fields were renamed. No CI run has happened: fork pull requests sit at `action_required` until a committer approves them, so all of the above is local only.

Two later commits close coverage gaps this work exposed. A cleaned file arrives in the metadata table as a `filesystemMetadata` entry with `size = 0, isDeleted = true`, which `HoodieMetadataPayload` writes as a constant and which `convertMetadataToFilesPartitionRecords` emits on clean and on rollback. Cleaning is on by default, so an ordinary table produces these, but no fixture here carried one because the fixture table was never cleaned: the fold that cancels such an entry ran against no data in either reader. Getting it wrong means a cleaned file either lingers in a listing or takes a live file out with it. The slice is built from two committed HFiles of twelve kilobytes each, a base file naming a file live and a log block naming the same file deleted, with the log framing assembled in the test rather than committed. Both readers return the partition key with an empty file map, so a tombstone cancels the entry; the test asserts they agree and asserts that outcome, and breaking version two's cancellation fails it.

The third commit covers the other case no fixture here can produce: a non-partitioned table, whose files partition holds one record keyed `.` and whose all-partitions record names that same partition `.` inside its file map. Both have to reach a caller as the empty string, and neither was exercised end to end, because no fixture table here is non-partitioned while carrying a metadata table; the normalisation was tested on constructed batches only. Three more synthetic HFiles cover them through both readers. The keyed read asks for `.` rather than the normalised form, since the predicate matches the key as stored, and the slice carries a base file and a log block naming different files so the union of the two file maps is asserted rather than just the key. Getting the all-partitions case wrong is the more expensive one: `partition_names` feeds the partition pruner, so a `.` left as-is prunes away the only partition a non-partitioned table has.

Two review findings from that pass are worth naming. The rationale I first wrote for using a base file plus a log block was false, and a mutation proved it: both the base side and the log side key the merge through `RecordContext::record_key_array`, so a normalisation there cannot desynchronise them, and the reasons the two-sided slice earns its keep are the ones stated above instead. The record type was also missing from what the two readers were compared on at this level, so a reader that mislabelled every record passed; it is compared now, and the metadata table's properties have one shared definition rather than a copy per test module. The fixtures' contents and how to regenerate them are stated in a README beside them.

Not covered, and stated rather than implied: the correctness fix makes a large block readable, not fast. Decode is superlinear in records per block, 6.72us each at 200 records against 68.89us at 5,000, which is a separate defect this fix made visible rather than caused. It is filed and not addressed here.


yihua:

PR #701, opened 2026-08-29
feat(core): stream file slices by partition

### LINKED ISSUE #590: Stream-style get_file_slices for large-table scalability
## Summary

`Table::get_file_slices` is fully eager today: it materializes the complete `Vec<FileSlice>` before returning. For tables with thousands of partitions or many file groups per partition, this produces noticeable startup latency and a large in-memory snapshot before any reads can begin. A streaming form would let engines start dispatching read tasks as slices are discovered.

## Current state

`Table::get_file_slices(options)` → `get_file_slices_internal(...)` → `FileSystemView::get_file_slices(...)` returns `Vec<FileSlice>` after:

1. Listing all partitions (filesystem walk OR metadata table query).
2. For each partition, listing all files.
3. Grouping files into file groups + slices.
4. Applying pruning (partition + file stats).

Every layer is eager. Engines that want bucketing call `hudi_core::util::collection::split_into_chunks` on the materialized vec — which still requires the full vec to be in memory first.

## Proposed shape

Add a streaming companion:

```rust
pub async fn get_file_slices_stream(&self, options: &ReadOptions)
    -> Result<BoxStream<'static, Result<FileSlice>>>
```

Per-partition granularity is the natural cut: stream partitions, materialize slices for each partition (still cheap — file groups per partition are bounded), apply pruning inline, emit. Partitions are the wide axis; per-partition slice materialization is fine.

## Layers that need streaming variants

| Layer                          | Today                     | Streamable form                                                  |
| ------------------------------ | ------------------------- | ---------------------------------------------------------------- |
| Object store listing           | `list()` returns `Vec<Path>` | `list_stream()` returns `Stream<Item = Path>` (object_store crate already supports this) |
| Partition discovery            | Build all paths upfront   | Yield partitions as listed                                       |
| File listing per partition     | List all files            | Yield per partition (cheap to buffer per-partition)              |
| Slice grouping                 | Group all files of a partition | Group within a partition; still buffer per-partition         |
| Pruning                        | Apply to materialized vec | Apply per slice as yielded                                       |
| Metadata table backend         | Returns full result set   | Stream rows from metadata table queries                          |

## Out of scope

- Per-slice streaming within a partition. The wide axis is partitions; slices per partition are bounded.
- Predicate pushdown changes — current pruning logic just runs inline in the stream.

## Use cases

- Large tables with thousands of partitions where startup latency on `get_file_slices` is observable.
- Engines that want pipelined planning (Ray Data, Spark with pipelined plans) that can dispatch task work before full planning completes.
- Memory-bounded environments where holding a 100k-slice vec is expensive.

## Notes

- The metadata-table-backed path needs the most refactor: `MetadataTable::get_file_slices` currently reads from another Hudi table; making it stream means streaming that read.
- Dispatch on `ReadOptions::query_type` as `get_file_slices` already does.
- Tests need to cover: streaming order, error propagation per-partition, empty-table handling, MDT-on/MDT-off parity.

goutamadwant:
## Description

Add `Table::get_file_slices_stream` so engines can begin scheduling reads before every partition and file slice has been planned.

For storage-backed tables, the new path:

- discovers partition paths from the object store's recursive stream;
- deduplicates and prunes partitions as paths arrive;
- bounds per-partition listings with `hoodie.plan.listing.parallelism`;
- applies footer pruning and timeline exclusion per partition;
- emits slices without a table-wide file-group or file-slice collection.

Snapshot and incremental query semantics match `get_file_slices`. Incremental planning still materializes the touched file-group set. Metadata-table-backed planning still materializes files-partition records because the metadata reader does not expose a row stream. Both paths use the new public stream API without changing the eager API.

The recursive discovery pass intentionally performs one delimiter listing for each emitted partition. Object stores do not guarantee partition-contiguous object order; reusing discovery entries would require retaining incomplete file groups table-wide. The stream instead retains one deduplication entry per emitted partition and bounds file groups and slices by listing concurrency. Global stream order is intentionally unspecified.

On a synthetic local table with 2,003 slices, the eager API returned only after all 2,003 slices were materialized. Post-change runs yielded the first slice in 244–262 ms and drained all 2,003 slices in 2.1–4.2 s under concurrent system load. These measurements demonstrate time-to-first-slice behavior, not a stable total-throughput improvement.

Closes #590.

Related: #597 optimizes the existing eager recursive listing path. This change adds a separate object-store-backed stream and does not modify `get_leaf_dirs`.

- [ ] N/A
- [x] Manual tests

Automated coverage verifies snapshot, incremental, metadata-table, and empty-table parity; first-slice-before-complete-discovery behavior; deferred listing-error propagation; recursive relative paths; and non-empty table-prefix boundaries.

Manual verification used the synthetic 2,003-slice table described above and confirmed that the eager and streamed APIs both returned all 2,003 slices.

Commands run:

- `cargo fmt --all -- --check`
- `cargo clippy -p hudi-core --all-targets --no-default-features -- -D warnings`
- `cargo clippy -p hudi-core --all-targets --all-features -- -D warnings`
- `cargo test -p hudi-core --no-default-features`
- `cargo test -p hudi-core --lib`
- `cargo check -p hudi -p hudi-cpp -p hudi-python --no-default-features`
- `cargo doc -p hudi-core --no-default-features --no-deps`

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ Create a Hudi table instance using its constructor or the `TableBuilder` API.
 
 All read APIs accept a `ReadOptions` (Rust) / `HudiReadOptions` (Python) value. It stores three fields — `filters`, `projection`, and `hudi_options` — and exposes chainable `with_*` builders for the rest. The available knobs:
 
-- `query_type` (`with_query_type`) — `Snapshot` (default) or `Incremental`. Drives dispatch in `read`, `read_stream`, and `get_file_slices`.
+- `query_type` (`with_query_type`) — `Snapshot` (default) or `Incremental`. Drives dispatch in `read`, `read_stream`, `get_file_slices`, and `get_file_slices_stream`.
 - `filters` — column filters as `(field, op, value)` tuples. The field can be any column (partition or data). Used for partition pruning, file-level stats pruning (snapshot only), and row-level filtering.
 - `projection` — columns to return. Streaming pushes the projection down to the parquet reader; eager reads project after merging.
 - `batch_size` (`with_batch_size`) — rows per batch (streaming only; eager reads return one batch per file slice).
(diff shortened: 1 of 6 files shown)


yihua (commented):
Hi @goutamadwant Thanks for your contribution! Could you rebase your PR on the latest main as the new file group reader implementation is landed?

PR #704, opened 2026-08-29
fix(core): keep the custom merger's own columns in the required schema

linliu-code:
## Description

The required schema keeps a merge's own inputs even when the caller did not ask for them — record key fields, ordering fields under event-time ordering, the delete markers, `_hoodie_operation`. It never consulted the **custom merger**, although `prepare_required_schema` had already resolved one into `self.custom_merger`.

Every metadata partition shares **one** Avro record. `HoodieMetadata.avsc` declares five nullable per-partition sub-records — `filesystemMetadata`, bloom filter, column stats, record index, secondary index — and exactly two non-nullable fields, `key` and `type`. They are non-nullable because they are the two every partition must carry: `key` identifies the record, and `type` is the discriminant that tells the merger which of the five rules applies, since the schema alone cannot distinguish them. Neither appeared in the lists above. So a requested schema that omitted them would have been honoured, the base read would have projected them away, and the merge would have failed with

```
A metadata record must carry a 'type' column to be merged.
```

on a perfectly valid metadata table.

**That refusal is a correct guard and is left exactly as it is.** What was wrong is that a projection could remove the column it guards.

`key` was already kept in practice — but only because the metadata table's config names it as the record key, so the record-key path picked it up. That is the config being right, not the merger's inputs being protected. Naming both columns on the merger makes them independent of it, and appending a name the list already holds is free: the caller only adds fields missing from the base schema, and deduplicates.

The fix puts the knowledge where it belongs: `CustomMerger` now declares the columns it reads, and the schema handler asks it, rather than the handler carrying names owned by the merger. A second custom merger added later cannot be forgotten at that one call site.

**Not reachable today, and the PR says so rather than overselling.** No production code sets a requested schema — every caller of `with_requested_schema` is under `#[cfg(test)]` (`schema_handler.rs`'s test module, plus `harness.rs` and `harness_tests.rs`, both `#[cfg(test)]` in `reader_v2/mod.rs`). With it unset, `generate_required_schema` falls back to the full data schema, which always carried `type`. The protection was accidental rather than designed, and the trigger is close: `hudi-datafusion`'s `TableProvider` already does projection pushdown at the table level, and the C ABI added for JVM callers exists specifically to let an outside caller drive reads. The first caller that threads a requested schema into a metadata read turns a correct guard into a spurious failure.

- [ ] N/A
- [ ] Manual tests

One test, `the_required_schema_keeps_the_custom_merger_s_own_columns`. The discriminating shape is a requested schema of `["_hoodie_commit_time"]` against a table schema of `["_hoodie_commit_time", "key", "type", "filesystemMetadata"]` on a `HoodieMetadataPayload` table. Three things make it discriminating:

- `key` and `type` are in the table and absent from the request, which is the only shape that separates *adding them back* from *never having dropped them*.
- **No record key fields are passed.** `key` is the metadata table's configured record key in practice, so passing it would let the record-key path keep the column and hide whether the merger's declaration does the work.
- `filesystemMetadata` is named by neither the request nor the merger, so it must stay projected away — without that check the test would pass if nothing were projected at all.

It also asserts its own premises (that a merger really resolved, and that the request really omits all three columns) so it cannot pass vacuously.

Mutation-checked with a positive control, on both halves of the fix:

| mutation | result |
| --- | --- |
| drop `key` from `required_field_names` | FAILED — required schema is `["_hoodie_commit_time", "type"]` |
| drop `type` from `required_field_names` | FAILED — required schema is `["_hoodie_commit_time", "key"]` |
| remove the call in `generate_required_schema` | FAILED — required schema is `["_hoodie_commit_time"]` |
| unmutated | passes |

Each column and the wiring fail independently, naming the column that went missing.

```
cargo test -p hudi-core --lib                                          1393 passed, 0 failed
cargo test -p hudi-core --lib --no-default-features                    1373 passed, 0 failed
```

**Not covered:** there is no end-to-end test reading a metadata table through a projection, because no production path can express one yet. When projection pushdown reaches this reader, that test becomes possible and should be added with it.


yihua (commented):
Thanks for working on this! With the stack consolidated here after the rebase onto latest main, leaving one review pass on the whole change: reader v2 gains HFile base files and log blocks on ordinary tables, the metadata payload's own merge rule, key-predicate seeks into the HFile block index, and the production `files`-partition metadata read moves onto the v2 reader. Both test legs pass locally on the rebased head (1400 / 1380, clippy and fmt clean), and the merge and read semantics were compared against the Java implementation in both 1.x and 0.x.

Three inline comments below, none blocking. Beyond those, here is the Java-parity delta from that comparison, recorded so it can be tracked as follow-ups in separate PRs rather than blocking this one.

**Missing — fine to land without, worth filing as follow-ups:**
- Valid-instant filtering of metadata log blocks (see the inline comment on `metadata/table/mod.rs`). One implementation of the 1.x rule covers table version 6 too — it subsumes the old 0.x suffix decoding (`001`–`006` / `>=010`).
- Key-to-file-group routing (`HoodieTableMetadataUtil.mapRecordKeyToFileGroupIndex`): only single-slice reads are served. Porting caveat: Java's hash is a wrapping 32-bit `31*h + charAt(i)` over UTF-16 units with a double `Math.abs` — a naive `(h as u32) % n` is not equivalent.
- Production reads serve the `files` partition only; col-stats / bloom / record-index / secondary-index point lookups (and SI's escaped `secKey$recordKey` encoding) are test-reachable but not wired.
- The bloom-filter meta-block pre-filter before keyed seeks, and the HFile block cache (`hoodie.hfile.block.cache.enabled`, default true in Java). Both perf-only.
- The spurious-delete check (`checkForSpuriousDeletes`) and `validatePayload`'s size>0 integrity check. Corrupt-input handling only.

**Divergent — deliberate or safe-side, recorded so each is a decision rather than a surprise:**
- Cross-type column-stat bounds are refused rather than promoted. On table version >= 9 Java promotes via `valueType`/`ValueMetadata`, so this is a real capability gap there (a bound merged under the wrong type would mis-prune files, so refusing is the right interim); on v6 Java itself throws `ClassCastException`.
- Keyed reads cap the whole-file read at 512 KB vs Java's single 50 MB threshold — I/O strategy only, justified by the absence of Java's file/block caches.
- `blocks_for_keys` takes the preceding block on an exact separator hit where Java takes only the floor block; HBase-written v6 files are not bound by the native writer's same-key-same-block rule, so the over-include is strictly safer, at one extra block read.
- Null counters treated as 0 (Java would NPE on unboxing); Java's double-delete warning log dropped.

This is one review pass, not an approval.

crates/core/src/metadata/table/mod.rs
```diff
@@ -272,7 +284,17 @@ impl Table {
...
+        // partition, measured by `v2_reader::tests::reader_cost_on_a_metadata_slice`,
+        // and accepted: roughly a quarter of that is `arrow_avro` re-parsing the
+        // writer schema's JSON on every decoder construction, which no change here
+        // can remove while a flushed decoder cannot be reused (arrow-rs#10876).
+        //
+        // The fixture is also this reader's worst case, holding 2.6 records per block,
+        // where per-block fixed cost cannot amortise; on larger blocks it overtakes
+        // the reader it replaces. So the ratio above is not the production ratio, and
+        // nothing measured here establishes what that is.
+        v2_reader::MetadataTableV2Reader::new(configs, storage)
```
yihua, line 297: non-blocking: Now that the production files-partition read goes through this layer, could the missing valid-instant filter get a tracked follow-up? Java gates every MDT log block on an EXACT_MATCH set from `getValidInstantTimestamps` (completed data-table instants + completed MDT deltacommits not pending on the data table + rollback-derived instants), so a data-table commit that fails after its MDT deltacommit completes is excluded there but admitted here — commit visibility catches it for file listing today, but nothing will for col-stats or RLI once they ride this layer.

crates/core/src/file_group/base_file/hfile.rs
```diff
@@ -0,0 +1,1587 @@
...
+                ))
+            })?
+            .to_string();
+        // The schema comes from the decoder, not from converting the Avro JSON:
+        // `avro_to_arrow` does not handle named-type references, and the metadata
+        // table's record schema uses them.
+        let registered = RegisteredWriterSchema::new(&json)
+            .map_err(|e| StorageError::Creation(format!("{e}")))?;
+        let decoder =
+            AvroBlockDecoder::try_new_with_registered(&registered, None, DECODE_BATCH_SIZE)
```
yihua, line 93: Has the HFile base path been exercised against an evolved schema? The values decode at the writer schema with no Avro writer-to-reader resolution (Java's `GenericDatumReader(writer, reader)` resolves), so an int-to-long promotion or added column relies entirely on the shared intersection/batch-evolution machinery — which looks right but has no HFile test behind it, and the missing schema-resolver arm this stack tracks separately is in the same territory.

crates/core/src/file_group/reader_v2/engine.rs
```diff
@@ -410,6 +415,13 @@ impl HoodieFileGroupReader {
         } else {
             BaseFileFormatValue::from_str(&self.reader_context.base_file_format)?
         };
+        // The shared factory refuses HFile, which is what keeps the legacy
```
yihua, line 418: non-blocking: this comment predates the factory change later in this stack — `create_base_file_reader` now serves HFile, so "the shared factory refuses HFile" is no longer true and this arm duplicates it. Could the special case just fall through to the factory, taking the stale comment with it?

yihua (approved):
LGTM overall.  Thanks for adding the new HFile and metadata table reader in Rust!

PR #706, opened 2026-08-29
feat(core): serve Hudi metadata tables through reader v2, sharded and memory-bounded

linliu-code:
## Description

Serves the Hudi metadata table through the v2 file-group reader, sharded and memory-bounded:

- **Metadata table on reader v2**: metadata partitions are read through `HoodieFileGroupReader`, with multi-slice discovery for any partition and record keys routed to their shard by a faithful port of Java's `String.hashCode` (UTF-16 code units, wrapping `i32`, `abs(abs(h) % n)`); the `files` partition is served end to end, and only the shards a key lookup routes to are opened.
- **Valid-instant set**: metadata log blocks are filtered by the instant set Java builds in `getValidInstantTimestamps` (completed deltacommits, the data timeline's pending set, and instants rolled back by rollback/restore), with rollback and restore now first-class timeline actions whose Avro metadata is readable, including the requested-plan fallback when a completed rollback file is empty.
- **Memory-bounded scans**: `hoodie.read.scan.max.memory.size` derives slices-in-flight from a per-slice cost model, applied by both the core `Table::read` path and the DataFusion plan; the plan also registers its reservation with DataFusion's memory pool.
- **fg-bench**: a benchmark harness for the file-group reader under `benchmark/filegroup/` (bounded-memory gate, spill detection, and the MOR table generator it needs).

The diff exceeds the usual size guideline because it consolidates the metadata-table-on-reader-v2 stack into one reviewable unit: its bases (#692, #700) have merged into main, and the remainder previously split across #702, #705, and #707 is carried here (those PRs can be closed).

- [ ] N/A
- [ ] Manual tests

Per-source tests on the valid-instant set (rollback, restore, pending-instant exclusion), a value-for-value parity test of the v2-backed metadata reader against the existing one, shard-routing vectors from Java's own `hashCode` output (including the `i32::MIN` edge case), DataFusion plan tests for budget-derived fan-out, a timeline test on the generated bench tables, and new metadata/MOR fixtures with documented generation scripts.


crates/core/src/metadata/table/mod.rs
```diff
@@ -294,9 +416,181 @@ impl Table {
...
+    }
+
+    /// Source 2 — completed metadata delta commits with no *pending* data
+    /// instant of the same name.
+    ///
+    /// An indexing delta commit has no data instant at all and is valid. One
+    /// whose data instant is still pending is the case where the data write
+    /// failed after the metadata write committed, and must not be read — that
+    /// exclusion is the whole content of this source.
+    fn valid_from_mdt_delta_commits(&self, mdt: &Table) -> HashSet<String> {
```
yihua: The exclusion here checks `pending_instants`, which is built from DEFAULT_LOADING_ACTIONS (commit/deltacommit/replacecommit), while Java builds `datasetPendingInstants` from `filterInflightsAndRequested()` over every action — so a completed MDT delta commit for a data compaction or clean that crashed mid-commit (a `.compaction.inflight` on the data timeline) is admitted here where Java excludes it. Today the files partition is re-gated downstream so this stays latent, but once column_stats/record_index gain consumers it becomes a silently wrong read — could we widen the pending listing for this check, or at least leave a comment naming the gap?

crates/core/src/metadata/table/mod.rs
```diff
@@ -294,9 +416,181 @@ impl Table {
...
+    ///
+    /// A rollback names them in its own metadata, falling back to its
+    /// `.requested` plan when the completed file is empty -- Java does the same
+    /// (`getRollbackedCommits`, HoodieTableMetadataUtil:2158). A restore is
+    /// several rollbacks, so its commits are the union of theirs.
+    ///
+    /// An unreadable instant yields nothing rather than failing the read: a
+    /// rollback we cannot parse should not make the whole table unreadable, and
+    /// the cost is the same conservative exclusion that existed before.
+    async fn commits_rolled_back_by(&self, instant: &Instant) -> Result<Vec<String>> {
```
yihua, line 580: Java's getRollbackedCommits throws HoodieMetadataException when a rollback/restore's metadata cannot be read, whereas this returns an empty list at debug level (the restore arm's unwrap_or_default likewise swallows a parse failure) — the read then silently serves listings missing the re-applied log blocks, which is wrong query results rather than a failed read. Would you consider propagating the error like Java, or at minimum a warn! so a corrupt rollback file is visible?

benchmark/filegroup/SKILL.md
```diff
@@ -0,0 +1,105 @@
...
+| ref | what | engine | in-memory format |
+|---|---|---|---|
+| `current` | hudi-rs reader (all A1–A6 fixes), `--streaming` | none (Rust) | Arrow columnar |
+| `java` | Java `HoodieFileGroupReader` via `HoodieAvroReaderContext` | none (pure JVM) | Avro row (`IndexedRecord`) |
+| `java-spark` | production columnar path via embedded `local[N]` Spark | in-process Spark driver | `InternalRow`/`ColumnarBatch` |
+
+`baseline` (pre-fix Rust binary) is a 4th optional ref if a baseline binary is staged.
+
+## Paths (verified 2026-06-11; the original `ws3`/`operations` paths are GONE)
+- **Repo**: `/home/ubuntu/lin_root/hudi-rs-internal` — branch `lin/mor-on-0607` already contains the A6 work (`benchmark/filegroup/` = the `fg-bench` crate).
```
yihua: This file (plus the ENG-XXXXX / milestone references in README.md, main.rs and the fixture docs) points at internal machines, branches and trackers — /home/ubuntu/lin_root/hudi-rs-internal, run_matrix.sh, ENG-42991 — none of which exist for apache/hudi-rs readers. Could you strip these down to what is reproducible from this repo (or drop SKILL.md entirely)?

benchmark/filegroup/src/main.rs
```diff
@@ -0,0 +1,734 @@
...
+                ALLOC_LIVE.fetch_sub(layout.size() as u64, Ordering::Relaxed);
+                // Null here, rather than a panic: a panic would unwind through
+                // an allocation path that has just been told there is no memory.
+                return std::ptr::null_mut();
+            }
+        }
+        unsafe { System.alloc(layout) }
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
```
yihua, line 212: Allocations made before ALLOC_CAP is set (tokio startup, the env string itself) never enter ALLOC_LIVE, but once the cap is set their frees are subtracted here, so the counter wraps and every later `live > cap` check fails — a capped run aborts with "memory allocation of 64 bytes failed" regardless of real usage (reproduced with a 512 MiB cap on the 36 MB fixture read). A saturating subtract via fetch_update, or counting unconditionally from process start, would fix it.

benchmark/filegroup/compare.py
```diff
@@ -0,0 +1,126 @@
...
+
+    row = {
+        "file": path,
+        "wall_ms": _median([it["wall_ms"] for it in measured]),
+        "max_rss_kb": max((it["max_rss_kb"] for it in measured), default=0),
+        "rows": measured[0]["rows"] if measured else 0,
+        "contended": report.get("contended", False),
+    }
+    for key, _ in STAGE_FIELDS:
+        row[key] = _median([it["read_stats"][key] for it in measured])
```
yihua: `it["read_stats"][key]` raises KeyError on fg-bench's own reports — IterationReport doesn't emit a read_stats object (and core's stats fields are _us-suffixed, not _ms), so the README's A/B workflow crashes on its first input. Could this be aligned with the JSON the binary actually prints?

benchmark/filegroup/src/main.rs
```diff
@@ -0,0 +1,734 @@
...
+    /// (`hoodie.read.scan.max.memory.size`). When set, slices-in-flight is
+    /// derived from it by `hudi_core::file_group::admission::slices_in_flight`
+    /// — the same function the DataFusion plan uses — and `--slice-concurrency`
+    /// becomes the ceiling it may not exceed rather than the value used.
+    #[arg(long)]
+    scan_memory_budget: Option<u64>,
+
+    /// Directory the merge map spills into (`hoodie.memory.spillable.map.path`).
+    /// Watched to report whether a run actually exercised the disk tier.
+    #[arg(long, default_value = "/tmp")]
```
yihua: With the default --spill-dir /tmp, the watcher reports the directory's absolute size, so any host with a non-empty /tmp reports spilled=true with a multi-GB spill_peak_bytes before the read does anything (I saw 48.5 GB) — and it re-walks all of /tmp every 100 ms. Defaulting to a fresh per-run subdir, or reporting growth over the initial size, would make the detector mean what it says.

crates/core/src/metadata/table/mod.rs
```diff
@@ -294,9 +416,181 @@ impl Table {
...
+        //
+        //    Only rollbacks newer than the earliest valid instant can have
+        //    rolled back anything we hold a log block for; Java bounds the scan
+        //    the same way, falling back to the sentinel when the set is empty.
+        let earliest = valid
+            .iter()
+            .min()
+            .cloned()
+            .unwrap_or_else(|| SOLO_COMMIT_TIMESTAMP.to_string());
+        for instant in self.rollback_and_restore_instants().await? {
```
yihua: non-blocking: Java bounds this scan by the later of the earliest valid instant and the latest MDT compaction time (rollbackFilterThreshold in getValidInstantTimestamps), specifically to avoid sequential reads of old rollback files during metadata reads — this loop reads every rollback newer than the earliest valid instant, one awaited GET at a time, per metadata read. Worth adopting the compaction bound; the comment above saying Java bounds it "the same way" would then also be accurate.

crates/core/src/table/fs_view.rs
```diff
@@ -278,17 +278,33 @@ impl FileSystemView {
...
         &self,
         partition_pruner: &PartitionPruner,
         file_pruner: &FilePruner,
         table_schema: &Schema,
         timeline_view: &TimelineView,
         metadata_table: Option<&Table>,
+        // The instants whose metadata log blocks may be read. Built by the data
+        // table, which holds both timelines, and passed in because this view
+        // holds only the metadata one.
+        valid_instants: Option<&std::collections::HashSet<String>>,
```
yihua: non-blocking: valid_instants=None with metadata_table=Some(..) silently falls back to an empty set, which admits no metadata log blocks at all — the strongest filtering, where None naturally reads as "no filtering". fs_view_get_file_slices_with_metadata_table already exercises that degraded path without being able to notice; bundling the metadata table and its valid set into one parameter would make the bad state unrepresentable.

benchmark/filegroup/src/main.rs
```diff
@@ -0,0 +1,734 @@
...
+            cfg.scan_memory_budget.unwrap_or(0) / (1024 * 1024),
+            file_slices.len(),
+            cfg.slice_concurrency,
+        );
+    }
+    let total_rows = futures::stream::iter(file_slices.iter())
+        .map(|slice| {
+            let reader = &reader;
+            let read_options = &read_options;
+            let streaming = cfg.streaming || cfg.async_stream;
```
yihua: non-blocking: --streaming and --async-stream document two different mechanisms (sync open() on OS threads vs open_blocking_stream()), but both reduce to the same async read_file_slice_stream call — a sync-vs-async comparison run with these flags measures the same path twice. Either wire the second mode or fold the flags into one; the README's stage-timings table has the same drift.

benchmark/filegroup/src/gen.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Rows per parquet row group. Smaller groups let a reader release memory
+    /// sooner, so this is the knob that makes a bounded read possible at all.
+    #[arg(long, default_value_t = 100_000)]
+    row_group_rows: usize,
+}
+
+const INSTANT: &str = "20250101000000000";
+/// The log files are written at a later instant than the base file, so their
+/// records win under commit-time ordering and the merge actually has work to do.
+const LOG_INSTANT: &str = "20250102000000000";
```
yihua, line 89: The generated log records carry an instant that never commits on the timeline, so they're only readable because the standalone FileGroupReader's uncommitted-block gate degrades to admit-all without timeline inputs — a Table::read of the same generated table would drop them. Should fg-gen also write the {instant}.deltacommit, or the README note the table is valid only through the standalone reader path?

crates/core/src/metadata/table/mod.rs
```diff
@@ -233,18 +271,104 @@ impl Table {
...
+        let batches = crate::util::concurrency::bounded_in_order(
+            &targets,
+            self.bounded_read_concurrency_for(&targets),
+            |file_slice| reader.read_files_partition_batch(file_slice, keys),
+        )
+        .await?;
+        concat_metadata_batches(batches)
+    }
+
+    /// Resolve the `files` partition's single file slice and build a reader for it.
```
yihua: non-blocking: a few editing leftovers in the docs — files_partition_reader carries the old "single file slice" doc on top of the new one, the NOTE below it names a partition_reader_with_valid_instants that doesn't exist, MetadataTableV2Reader::read_files_partition has its doc block twice, bounded_read_concurrency_for's doc has stray fragments, and read_file_slices_bounded still links [`Self::file_slice_read_concurrency`], which is gone. Worth one cleanup pass, since these will actively mislead the next reader.

crates/core/src/table/mod.rs
```diff
@@ -842,25 +851,67 @@ impl Table {
...
+    /// call whose absence compiles perfectly, and the arithmetic behind it is
+    /// tested elsewhere in isolation, which says nothing about whether anything
+    /// calls it.
+    /// [`Self::bounded_read_concurrency`] for a borrowed selection.
+    ///
+    /// The metadata read routes keys to a subset of a partition's slices, so it
+    /// holds `&FileSlice` rather than owning them. The budget must see that
+    /// subset, not the whole partition: routing a single key to one shard should
+    /// admit on that shard's cost.
+    pub(crate) fn bounded_read_concurrency_for(&self, file_slices: &[&FileSlice]) -> usize {
```
yihua, line 905: nit: this clones every routed slice only so bounded_read_concurrency can take &[FileSlice]; computing the log_size_bytes vector directly from the borrows would avoid the copies.

crates/datafusion/src/hudi_exec.rs
```diff
@@ -293,11 +317,29 @@ impl ExecutionPlan for HudiScanExec {
...
+        // the pool is already under pressure, read fewer slices at once instead
+        // of allocating anyway. Dropping the reservation with the stream returns
+        // the bytes.
+        let mut reservation = MemoryConsumer::new(format!("HudiScanExec[{partition}]"))
+            .register(context.memory_pool());
+        let per_slice = file_slices
+            .iter()
+            .map(|s| {
+                hudi_core::file_group::admission::estimated_slice_bytes(
+                    s.log_size_bytes().unwrap_or(0),
```
yihua, line 347: non-blocking: slices_in_flight treats an unrecorded log size as "cannot estimate, admit one", but this reservation treats it as zero bytes — an unmeasured log-heavy slice reserves only the 33 MiB floor, understating pressure on the pool. Using the same conservative rule in both places would keep the estimate's meaning consistent.

yihua (commented):
Thanks for working on this — and for restructuring the branch onto current main; the +5.5k diff now reads as one coherent change: the metadata table served through reader v2 with shard routing, the five-source valid-instant set, rollback/restore as first-class timeline actions, and a scan-wide memory budget shared by the core and DataFusion paths.

The core work here is very solid — the Java `hashCode` port with its UTF-16 and `i32::MIN` edge cases, the mutation-hardened per-source tests on the valid-instant set, and the measured admission coefficients are all careful in ways that are easy to get wrong. The inline comments cluster around three themes: two places where the valid-instant set diverges from Java's semantics (the pending-set scope for source 2, and error handling on unreadable rollback metadata); the benchmark harness, which ships with a broken allocation cap, a compare script that crashes on the harness's own output, and a spill detector that always fires on the default directory; and internal ENG-/machine references that need to come out before this can land in an ASF repo. The PR description also predates the rebase (it still describes 28 commits / +17.4k and the old #702-based composition) and is worth refreshing.

This is one review pass, not an approval — I'd like another look after the valid-instant-set questions are settled.

crates/core/src/metadata/table/mod.rs
linliu-code: Confirmed, and widened rather than commented. `Timeline::pending_instants` is built from `DEFAULT_LOADING_ACTIONS`, and the `Action` enum has no variant for compaction/clean/cluster at all — `Action::from_str` rejects them, so a selector-based listing drops those files before anything can count them.

`Timeline::all_pending_instant_times` now lists the active timeline and parses instant times structurally (state from the suffix, instant time from what precedes the first `.`, minus the `_completionTime` layout 2 appends). That matches what Java actually needs from `filterInflightsAndRequested()` — the timestamps, never the actions.

Deliberately a separate set, not a widening of `pending_instants`: that field also feeds `completion_gate_inputs().inflight_instants`, so widening it in place would start rejecting log blocks the MOR completion gate admits today. Only source 2 takes the wide view.

Test is `source_two_excludes_a_data_instant_pending_under_a_non_commit_action`: a metadata delta commit at an instant whose only data-side file is `{ts}.compaction.inflight`. It asserts the wide set sees it, that `pending_instants` does *not* (pinning the gap itself), and that the instant is excluded. Plus two loader unit tests covering both timeline layouts and the completed-cancels-pending subtraction.

crates/core/src/metadata/table/mod.rs
linliu-code: Propagated, matching Java. `commits_rolled_back_by` now returns `CoreError::MetadataTable` — the counterpart to `HoodieMetadataException` — when the instant will not load, when a restore's metadata will not parse, and when the rollback *plan* will not load or parse.

The one fallback Java keeps is kept: an unreadable or empty completed rollback file falls back to its `.requested` plan, and that now logs at `warn!` (Java's `LOG.warn`) instead of passing in silence. A plan naming no instant stays a non-error — that is a rollback of nothing, not a failure.

Your framing is the reason: the swallow dropped exactly the commits rollback re-applied, so the listing came back missing their log blocks — a wrong answer presented as a right one, which is worse than a read that stops.

Regression test `an_unreadable_rollback_fails_the_read_rather_than_shrinking_the_set` writes a corrupt completed rollback with no plan beside it and asserts the error names the instant.

benchmark/filegroup/SKILL.md
linliu-code: Dropped SKILL.md entirely — it was wholly about one internal host. Also stripped the `ENG-` ids, the `A1`/`A3`/`A6e` milestone labels, and the `perf M2` / `02-bench-harness-design.md` / `03-milestones.md` references from README.md and main.rs.

The README needed more than a trim: its "Stage timings" table and the `read_stats` field documented an interface the harness lost when it moved to the public reader surface. Rewritten to describe what the binary actually emits and what is reproducible from this repo alone.

benchmark/filegroup/src/main.rs
linliu-code: Good catch, and reproduced. Minimal repro of the mechanism:

```
cap installed; LIVE=0
after freeing pre-cap alloc; LIVE=18446744073701163072
now allocating 1 KiB under a 512 MiB cap...
memory allocation of 1024 bytes failed
memory allocation of 64 bytes failed
```

`#[tokio::main]` builds the runtime before the body runs, so those allocations — and the env string itself, dropped at the end of the `if let` *after* the store — are never added but are subtracted once the cap is set.

Went with counting unconditionally from the process's first allocation rather than a saturating subtract: with balanced accounting the counter cannot underflow, so no CAS loop on the dealloc path. The cap is still only *checked* when set.

Worth noting the severity is data-dependent, which is probably why it survived some runs — `fetch_add` wraps, so a small underflow is healed by the next allocation larger than it. A multi-MB one is not, and that is your 64-byte abort.

benchmark/filegroup/compare.py
linliu-code: Right on both counts — `IterationReport` has no `read_stats` field, and the stage timings it named went away when the harness moved to the public reader surface (`HoodieReadStats` is not on it). README.md:84 was advertising the same phantom object.

`compare.py` now compares what the binary emits: `wall_ms`, `user_ms`, `sys_ms` and `spill_peak_bytes` by median, `max_rss_kb` by peak (it is a high-water mark, so the max across iterations is the number a bound has to hold against), plus rows/spilled/contended on the footer. Reads use `.get(key, 0)` so a report from an older build still compares instead of raising.

Verified end to end against reports the binary actually wrote.

benchmark/filegroup/src/main.rs
linliu-code: Fixed both halves.

`--spill-dir` now defaults to a fresh per-run subdirectory of the system temp dir (`fg-bench-spill-<pid>`), which also keeps the 100 ms sampler off large unrelated trees. And `SpillWatcher` samples the directory's size when it starts and reports `peak - baseline`, saturating — so growth, not absolute size, even when `--spill-dir` is pointed somewhere shared.

Checked against a directory pre-filled with 200 MB: `spilled=false, peak=0` both by default and when explicitly pointed at it. Confirmed it still detects a real spill (`--merge-max-size 65536` → `spilled=true, peak=27718`), so the fix did not trade the false positive for a false negative.

crates/core/src/metadata/table/mod.rs
linliu-code: Checked this against the source and I think the premise is off — `rollbackFilterThreshold` is not in apache/hudi.

apache/hudi `HoodieTableMetadataUtil.getValidInstantTimestamps`:2101-2104 bounds the scan on `earliestInstantTime` alone:

```java
final String earliestInstantTime = validInstantTimestamps.isEmpty() ? SOLO_COMMIT_TIMESTAMP : Collections.min(validInstantTimestamps);
datasetTimeline.getRollbackAndRestoreTimeline().filterCompletedInstants().getInstantsAsStream()
        .filter(instant -> compareTimestamps(instant.requestedTime(), GREATER_THAN, earliestInstantTime))
```

`grep -rn rollbackFilterThreshold` over apache/hudi master returns nothing; the only hit is in the Onehouse fork (`hudi-internal` master, the 0.14.1 line), where it was added as a latency optimization. So the existing "Java bounds the scan the same way" comment is accurate for the Java this project targets, and adopting the compaction bound here would be importing fork-only behavior — and narrowing the set on an argument about MDT compaction semantics I would not want to land unverified in an ASF repo. Happy to do it as its own change if you would rather hudi-rs lead apache/hudi here.

The cost you identified is real regardless of the bound, so I fixed that part: the reads are no longer a sequential awaited-GET walk. They now run through `bounded_in_order` at a concurrency of 8. Same set, off the critical path. I also added the `:2101` line reference to the comment so the claim is checkable.

crates/core/src/table/fs_view.rs
linliu-code: Agreed, and made unrepresentable. The two parameters are now one:

```rust
pub(crate) struct MetadataListing<'a> {
    pub table: &'a Table,
    pub valid_instants: &'a HashSet<String>,
}
```

`get_file_slices` takes `Option<MetadataListing<'_>>`, so there is no way to pass a metadata table without its set and no `unwrap_or(&empty)` left. That also drops the parameter count back under clippy's limit, so the `#[allow(too_many_arguments)]` and the comment explaining it are gone.

`fs_view_get_file_slices_with_metadata_table` — the test you noted was riding the degraded path — now builds the real set and asserts it is non-empty first, so it cannot pass while silently reading a metadata table it had filtered down to nothing.

benchmark/filegroup/src/main.rs
linliu-code: Correct — `let streaming = cfg.streaming || cfg.async_stream;` was the whole of it, and the docs described `open()` and `open_blocking_stream()`, neither of which exists on the public `FileGroupReader` (it has one streaming entry point, `read_file_slice_stream`).

Folded into a single `--streaming`, with the doc saying what it actually contrasts: streamed batches dropped as counted, versus the eager read retaining the whole output. The README's stage-timings table went too — it documented per-stage fields no report carries.

benchmark/filegroup/src/gen.rs
linliu-code: Confirmed, with numbers, and fixed in the generator. On a table from `fg-gen --files 2 --log-files 2 --log-records 5000 --log-key-offset 50000000`:

- `Table::read` with the delta commit: **90,000** rows
- same table, delta commit removed: **80,000** rows

So the 10,000 log records were being silently dropped on the `Table` path while the standalone reader returned all of them — measurements taken on such a table would have been comparing two different row sets.

`fg-gen` now writes `{LOG_INSTANT}.deltacommit` whenever it writes log files. Its write stats carry `baseFile` alongside `path`, because schema resolution reads the first write stat's `path`, gets nothing from a `.log.` extension, and falls through to the base file — without it the generated table loses its resolvable schema once the delta commit becomes the latest.

Regression test `benchmark/filegroup/tests/fg_gen_timeline.rs` runs the generator and asserts the delta commit exists for a MOR table and does not for `--log-files 0`.

crates/core/src/metadata/table/mod.rs
linliu-code: All five fixed:

- `files_partition_reader` — the superseded single-slice block above the current one is gone; one doc, and it says "every file slice" rather than "single file slice".
- the NOTE — rewritten to name `MetadataTableV2Reader::with_valid_instants`, which is what actually attaches the set, and to say why this constructor cannot (it holds only the metadata table's timeline).
- `read_files_partition` — the duplicated paragraph merged.
- `bounded_read_concurrency_for` — its doc had been spliced into the middle of `bounded_read_concurrency`'s; the two are separated again.
- `read_file_slices_bounded` — now links `Self::bounded_read_concurrency`.

Checked `cargo doc` before and after: 13 unresolved links both ways, all pre-existing in files this PR does not touch.

crates/core/src/table/mod.rs
linliu-code: Done. Both forms now funnel through `bounded_read_concurrency_from_log_bytes`, which takes an `Iterator<Item = Option<u64>>`, so each derives the log-size vector from its own borrows and the owned round-trip is gone. Admission needs nothing from a slice except its log size, so neither caller has to own one.

crates/datafusion/src/hudi_exec.rs
linliu-code: Agreed — the two halves of the same budget disagreed about what an unmeasured slice costs. Extracted the rule into `HudiScanExec::planned_slices`, which applies `slices_in_flight`'s: any slice with an unrecorded log size drops the plan to one, instead of letting `unwrap_or(0)` reserve just the 33 MiB floor for the slices least is known about.

Test `an_unmeasured_log_file_plans_one_slice` pins them together — 2 slices with known log sizes plan 2 under a ceiling of 4; adding one unmeasured slice drops it to 1.

benchmark/filegroup/SKILL.md
yihua: Done. Rebased on latest main and stripped the internal references: SKILL.md is dropped entirely, and the ENG/milestone/codename references are gone from the bench README and CLI help, the v2 reader test comment, and both fixture docs. The few commit messages that pointed at internal branches and tickets are reworded too. The old strings still exist in intermediate commits of the branch, so this should land as a squash merge.

yihua:
Heads up @linliu-code: we were both pushing this branch today, so I force-pushed once more on top of your 18:49 rebase. Your three review-response commits are preserved on the new head; the only additions are one internal-reference cleanup commit and reworded messages on the four commits that mentioned internal branches or tickets. Please hard-reset your local branch to the remote before further pushes.

yihua (approved):
LGTM

crates/core/src/metadata/table/mod.rs
yihua: Re-checked this against current upstream, and the premise holds: rollbackFilterThreshold is in apache/hudi master, added by apache/hudi#18544 (89f49b18d0af, 2026-05-19) in getValidInstantTimestamps, for exactly this latency concern. The line numbers quoted above match a pre-18544 checkout, which is where the "not in OSS" conclusion came from; the in-code comment here claiming Java bounds the scan the same way is likewise stale against upstream.

One sharper version of the concern: sentinel commits (00000...) enter the valid set before the min(), so on a table whose MDT has a sentinel deltacommit the earliest-instant bound degenerates to about zero and every completed rollback/restore is read on each metadata read. The bounded fan-out of 8 caps concurrency, not the count.

Still non-blocking for this PR; tracking the threshold port and the sentinel fix as a follow-up.

benchmark/filegroup/SKILL.md
yihua: Pushed 847cce96 on top with the few that remained: the ENG ids in the v2 reader test comment and the two fixture docs, and the M2/M4/fork mentions in the bench crate. A few intermediate commit messages still cite internal ids, so this should land as a squash merge, which keeps them out of main.

PR #707, opened 2026-08-30
feat(core): serve sharded metadata partitions — routing, instant set, rollback-aware timeline

linliu-code:
## Description

Five increments of ENG-47771. The first three are below; the last two are summarised here because they are the substance of the change:

**4. The timeline can see rollback and restore instants.** `Action` had three variants and the parser rejected everything else — and the loader *discards* a parse failure at debug level, so a `.rollback` file made a table open fine with those instants invisible. Two of the valid-instant set's five sources are made entirely of them. Deliberately **not** added to `DEFAULT_LOADING_ACTIONS`, which fills `completed_commits` and would change every existing read.

**5. The valid-instant set, assembled and wired.** Five sources ported from `HoodieTableMetadataUtil.getValidInstantTimestamps` (:2081), with readers for `HoodieRollbackMetadata`, `HoodieRestoreMetadata` and `HoodieRollbackPlan` — the last being Java's fallback when the completed rollback file is empty. `InstantRange::exact_match` gets its first caller and the metadata read stops using a window.

Two things worth a reviewer's attention in increment 5:

- **The first test for the set could not fail.** It asserted membership in the *union*, and three mutations — one per source — all passed. Sources 1, 2 and 5 overlap almost entirely, because a data commit and the metadata delta commit recording it share a timestamp, so every member survives dropping any single source. Each source is now its own function, tested on its own output; the same three mutations all fail. The sources 3 and 4 test uses the other technique — an input only one source can supply, asserted absent from the others first.
- **Clippy found a real defect while wiring:** `read_files_partition_batch` took the set and never used it, so the Arrow path would have kept the window while the decoded path used the set. Two read paths silently disagreeing about which log blocks are readable.

Verified by mutation rather than by the suite passing: making the set come back empty fails six tests. A set that is merely *wrong* rather than empty would still let most reads succeed, so a green suite proves the reads work, not that the filter does anything.

```
cargo test -p hudi-core --lib                          1429 passed, 0 failed
cargo test -p hudi-core --lib --no-default-features    1409 passed, 0 failed
cargo test --workspace --all-targets --all-features    0 failing suites
clippy (both feature sets) · fmt                       clean
```

**Not covered, stated rather than implied.** Restore is untested end to end — source 3 handles rollbacks *and* restores, and only the rollback branch has a synthetic instant; `RestoreMetadata`'s reader is tested on its own. The empty-file fallback to the `.requested` plan is likewise reader-tested only. And the set is recomputed on every `get_file_slices` call, walking both timelines and reading rollback metadata from storage; Java caches it on the metadata client, nothing here does.

**Remaining in ENG-47771 and not in this PR:** a decoded record type per partition. That is blocked on fixture tickets ENG-47731 and ENG-47745 — the current fixture structurally cannot exercise five of six merge rules, and every Avro-carrying HFile in the repo has exactly one data block.

---

## 1. Shard routing

---

## 1. Shard routing With a metadata partition now readable across several file slices (#705), a key lookup still opened **every** slice. A partition that shards — `record_index` has ten file groups in the fixture, `secondary_index_rider_idx` ten — hashes each key to exactly one, so nine of those ten reads return nothing.

**The routing is cherry-picked from `wip/mdt-file-group-routing`, not rewritten**, as the ticket asks. It ports Java's `String.hashCode` over UTF-16 code units with wrapping `i32` arithmetic, then `abs(abs(h) % n)`, with vectors taken from Java's published values — including `polygenelubricants`, whose hash is exactly `i32::MIN` and is the only input the doubled `abs` exists for. Two traps it avoids are silent rather than loud: iterating Rust `char`s hashes a non-BMP character as its scalar and routes it to a different file group, and computing in `i64` diverges once the hash exceeds `i32`.

`slices_for_keys` turns that hash into a selection:

- **Slices are sorted by file id first.** `file_group_index` returns a position *among the shards*, while the listing order is the storage's. Hudi embeds the shard number in the file id (`record-index-0003-0`), so a lexicographic sort recovers it. Indexing an unsorted list sends a key to the wrong shard, which returns **no rows rather than an error** — the failure this sort exists to prevent.
- **Shards are deduplicated**, so a repeated key does not open its shard twice.
- **An empty key set still opens every slice.** That is not an omission: a prefix lookup cannot know which shard holds a match, because sharding is by the *full* key. Java says the same at `getRecordsByKeyPrefixes:239`.

Concurrency is unchanged — the selected slices go through the same bounded fan-out from #705, on the same `hoodie.read.file.slice.read.concurrency`.

- [ ] N/A
- [ ] Manual tests

**Fan-out width is asserted as slices selected, not inferred from rows returned**, which is what the ticket specifically asks for — and the reason matters: a wrong shard and an empty shard both return no rows, so counting rows cannot tell them apart.

Four tests over ten synthetic shards:

| test | what it pins |
| --- | --- |
| `a_key_lookup_opens_only_the_shards_its_keys_route_to` | one key selects one slice, **and it is the shard the hash names** — not merely some shard |
| `distinct_shards_are_opened_once_each` | two keys on different shards select two; the same key twice selects one |
| `an_empty_key_set_opens_every_slice` | a scan opens all ten |
| `selection_does_not_depend_on_listing_order` | reversing the input does not change the selected shard |

The second test asks the routing function which keys land on different shards rather than hard-coding a hash, so it is not checking the implementation against itself.

**Mutation-checked with a positive control**, each caught by a different test:

| mutation | result |
| --- | --- |
| drop the file-id sort | FAILED (1 test) |
| drop the shard deduplication | FAILED (1 test) |
| always scan every slice | FAILED (3 tests) |
| unmutated | 8 passed |

```
cargo test -p hudi-core --lib                                          1413 passed, 0 failed
cargo test -p hudi-core --lib --no-default-features                    1393 passed, 0 failed
```

**Not covered, stated rather than implied.** These four tests use synthetic `FileSlice`s. Routing against the fixture's **real** shards is covered by increment 3 below; what remains uncovered is which *keys* each shard holds, which needs a decoded `record_index` record type.

---

## 2. An exact-match `InstantRange`

The metadata table filters its log blocks by **set membership**, not by a window. That set has **holes** — a pending data instant is excluded while instants on either side are included — and members from outside the data timeline entirely, such as metadata-only indexing delta commits. `InstantRange` could express only bounds, and no bound admitting both neighbours can exclude what sits between them.

Why this matters beyond tidiness: for `files` a window is safe, because every listed file is gated downstream on `CompletionTimeView::is_committed`, so a file named by an uncommitted block is dropped anyway. **The other five partitions have no such gate.** A `column_stats` record drives pruning directly, so a record from a block written for a still-pending instant would be used as-is — and wrong statistics prune away files that hold matching rows, which is a *silently wrong query result* rather than a failure.

`InstantRange::exact_match` mirrors Java's `RangeType::EXACT_MATCH`: when the set is present, bounds are ignored entirely and membership is the whole test. Both predicates consult it, since the log-block gate falls back to the lexicographic one for instants that will not parse, and the two disagreeing would make the filter depend on whether a timestamp happened to be parseable.

### Tests

The test that matters is the one a window **cannot** pass: a set admitting two instants and excluding the one between them — paired with a control showing the equivalent window really does admit that middle instant. Without the control the assertion would be true for reasons unrelated to the change.

Building that control found its own bug: the window's start was written `20250100000000000`, a day zero, which failed to parse. The fixture was wrong, not the assertion, and it failed loudly rather than quietly passing.

| mutation | result |
| --- | --- |
| `is_in_range` ignores the set | FAILED (3 tests) |
| the lexicographic predicate ignores the set | FAILED (2 tests) |
| an empty set admits everything | FAILED (1 test) |
| unmutated | 5 passed |

The last matters because a filter handed no instants that admitted everything would be a no-op wearing the shape of a filter.

```
cargo test -p hudi-core --lib                          1417 passed, 0 failed
cargo test -p hudi-core --lib --no-default-features    1397 passed, 0 failed
cargo test --workspace --all-targets --all-features    0 failing suites
clippy (both feature sets) · fmt                       clean
```

**Nothing builds the valid-instant set yet.** That is the metadata layer's next piece; this exists so it has somewhere to put it.

---

## 3. Unpinning slice discovery

Discovery was pinned to `files` by a filter written into the reader. Nothing about listing slices depends on which partition is listed — the partitions that shard differ only in how many come back and how their records decode. `partition_reader` takes the partition name; `files_partition_reader` becomes a thin wrapper.

**This does not widen what is publicly served.** `files` remains the only partition with a decoded record type and the only one with a caller. What the parameter buys now is a test that could not be written before: shard routing checked against the fixture's **real** record index — ten file groups, written by Hudi, with Hudi's own file-id naming.

That matters because increment 1's tests use synthetic slices. They prove the selection logic, but not that the sort recovering shard order survives real ids like `record-index-0003-0`. It does — and the mutation shows the difference is not theoretical:

| mutation | result |
| --- | --- |
| drop the file-id sort | FAILED — selects `record-index-0004-0` where the hash names `record-index-0002-0` |
| restore the partition pin | FAILED on the premise — 1 slice where the fixture has 10 |
| unmutated | passes |

A wrong shard is a shard that exists and holds records, just not this key. A lookup there returns nothing — indistinguishable from a key that was never written, which is why the assertion names the expected shard rather than checking that *some* shard was selected.

The test states what it does not do: `record_index` has no decoded record type yet, so which keys each shard holds is unknown here. It pins discovery and selection, not decoding.

```
cargo test -p hudi-core --lib                          1418 passed, 0 failed
cargo test -p hudi-core --lib --no-default-features    1398 passed, 0 failed
cargo test --workspace --all-targets --all-features    0 failing suites
clippy (both feature sets) · fmt                       clean
```

---

**Remaining in ENG-47771**, none of it in this PR: a decoded record type per partition (five of six have none), building the valid-instant set from its five sources and handing it to the log-block gate, and comparing each partition against an oracle that is not this code.


yihua:
Closing this PR since the set of changes has been reviewed in #707

PR #711, opened 2026-09-03
feat(deps): upgrade to arrow 58, DataFusion 54, object_store 0.13, and Lance 11

### LINKED ISSUE #710: Upgrade arrow-rs, object_store, DataFusion, Lance, and pyo3
The workspace is several majors behind on its Arrow-side dependencies: arrow 57, parquet 57, object_store 0.12, DataFusion 52, pyo3 0.26, and Lance 4.0.1 (Lance is now at 11.0.0).

Lance sets the ceiling for how far this can go. Every published `lance-*` release requires `arrow ^58`, and so does lance's unreleased `main` (12.0.0-beta), so arrow 58 is the highest the workspace can take while keeping the Lance base-file reader compiling. That in turn caps DataFusion at 54, which requires arrow 58.3.

So this issue covers the move to arrow 58 / parquet 58 / object_store 0.13 / DataFusion 54 / Lance 11, plus the pyo3 0.28 that `arrow/pyarrow` forces on the way. DataFusion 55 needs arrow 59 and is out of reach until Lance publishes an arrow-59 release; datafusion-python is also still on 54, so the `hudi[datafusion]` FFI extra has no matching pin at 55 either. That is what blocks #698.

yihua:
## Description

closes #710

Moves the workspace to arrow 58.4, parquet 58.4, object_store 0.13.2, DataFusion 54.1, Lance 11, and pyo3 0.28. Lance is what sets the ceiling: every published lance release, and its unreleased `main`, requires `arrow ^58`, so 58 is the highest arrow this repo can take while the Lance base-file reader still compiles, and DataFusion 54 follows from arrow 58.

Most of the churn is mechanical (object_store 0.13 moved `head`/`get`/`copy` onto `ObjectStoreExt` and made `delete_stream`/`copy_opts` required; DataFusion 54 dropped `as_any` and returns `Arc` from `properties`/`partition_statistics`; Lance 11 moved `ReaderProjection::from_column_names` to a free function and made `read_stream` async; pyo3 0.28 replaced the capsule name-then-read pair with `pointer_checked`). Two things are worth a reviewer's attention: DataFusion 54 now pushes the projection and the struct-field predicate into the Parquet source rather than planning a `ProjectionExec` and a `FilterExec`, so the plan assertions in `plan_tests` were rewritten against the new shape (they assert the same projection, filter, and file-group count, just where the plan puts them now); and `benchmark/tpch` cannot follow the workspace, because tpchgen-arrow ships an arrow-57 line and an arrow-59 line and nothing on 58, so its self-contained generator builds against its own renamed arrow and parquet.

DataFusion 55 stays out of reach until Lance publishes an arrow-59 release and datafusion-python catches up (54.0.0 is the newest, on datafusion-ffi 54), which is what #698 is blocked on.

### DIFF
--- a/crates/datafusion/tests/plan_tests.rs
+++ b/crates/datafusion/tests/plan_tests.rs
@@ -149,34 +149,43 @@ async fn verify_plan(
         "Plan should contain TopK sort"
     );
     // The projection, struct field access included, is pushed into the Parquet
-    // source rather than planned as its own `ProjectionExec`, so it is asserted
-    // wherever it lands. The index in `structField@N` is the field's position in
-    // the file schema and differs per table, hence the two-part match.
+    // source rather than planned as its own `ProjectionExec`. Keeping the whole
+    // column list in the anchor is what makes this catch a scan that stopped
+    // projecting; only the `structField@N` index varies per table, so the alias
+    // is matched separately.
     assert!(
-        plan.contains("get_field(structField@")
+        plan.contains("projection=[id, name, isActive, get_field(structField@")
             && plan.contains(&format!(", field2) as {table_name}.structField[field2]")),
         "Plan should project the struct field"
     );
     // Simple predicates (id % 2 = 0, name != Alice) and the struct field access
     // alike are pushed into the Parquet source.
     assert!(
-        plan.contains(", field2) > 30"),
-        "Plan should contain struct field filter"
+        scan_predicate(&plan).contains(", field2) > 30"),
+        "Scan predicate should contain the struct field filter"
     );
+    // `hoodie.read.input.partitions` decides how many groups the scan is split
+    // into. One group renders in the singular, which this prefix covers too.
     assert!(
-        plan.contains(&format!(
-            "file_groups={{{planned_input_partitioned} {}:",
-            file_group_noun(*planned_input_partitioned)
-        )),
+        plan.contains(&format!("file_groups={{{planned_input_partitioned} group")),
         "Plan should scan {planned_input_partitioned} file group(s)"
     );
 }
 
-/// How the plan spells the file-group count: `hoodie.read.input.partitions`
-/// decides how many groups the scan is split into, and DataFusion renders one
-/// group in the singular.
-fn file_group_noun(count: i32) -> &'static str {
-    if count == 1 { "group" } else { "groups" }
+/// The predicate the scan evaluates per row, which it renders as `, predicate=`.
+///
+/// The scan also renders a `, pruning_predicate=` derived from that one, and a
+/// bare substring search cannot tell the two apart. Only the former says what
+/// actually reached the source.
+fn scan_predicate(plan: &str) -> &str {
+    let predicate = plan
+        .split(", predicate=")
+        .nth(1)
+        .expect("plan should carry a scan predicate");
+    predicate
+        .split(", pruning_predicate=")
+        .next()
+        .unwrap_or(predicate)
 }
 
 async fn verify_data(ctx: &SessionContext, sql: &str, table_name: &str) {
@@ -400,7 +409,7 @@ mod v8_tests {
             "Should scan through DataSourceExec"
         );
         assert!(
-            plan.contains(", field2) > 30"),
+            scan_predicate(&plan).contains(", field2) > 30"),
             "Non-partition filters should reach the source"
         );
         assert!(
--- a/crates/datafusion/tests/plan_tests.rs
+++ b/crates/datafusion/tests/plan_tests.rs
@@ -148,24 +148,35 @@ async fn verify_plan(
         plan.contains("SortExec: TopK(fetch=10)"),
         "Plan should contain TopK sort"
     );
+    // The projection, struct field access included, is pushed into the Parquet
+    // source rather than planned as its own `ProjectionExec`, so it is asserted
+    // wherever it lands. The index in `structField@N` is the field's position in
+    // the file schema and differs per table, hence the two-part match.
+    assert!(
+        plan.contains("get_field(structField@")
+            && plan.contains(&format!(", field2) as {table_name}.structField[field2]")),
+        "Plan should project the struct field"
+    );
+    // Simple predicates (id % 2 = 0, name != Alice) and the struct field access
+    // alike are pushed into the Parquet source.
+    assert!(
+        plan.contains(", field2) > 30"),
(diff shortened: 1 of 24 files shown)


linliu-code (approved):

crates/datafusion/tests/plan_tests.rs
```diff
@@ -148,26 +148,37 @@ async fn verify_plan(
...
+    // source rather than planned as its own `ProjectionExec`, so it is asserted
+    // wherever it lands. The index in `structField@N` is the field's position in
+    // the file schema and differs per table, hence the two-part match.
     assert!(
-        plan.contains(&format!(
-            "ProjectionExec: expr=[id@0 as id, name@1 as name, isActive@2 as isActive, \
-            get_field(structField@3, field2) as {table_name}.structField[field2]]"
-        )),
-        "Plan should contain expected projection"
+        plan.contains("get_field(structField@")
```
linliu-code: This drops `id, name, isActive` from the assertion, so a projection-pushdown regression that read every file column would still pass here — `verify_data` only sees the query's output columns, not what the scan read, and `get_field(structField@` also appears inside `predicate=`. Since only the `@N` index varies per table, could the first half keep the anchor, `plan.contains("projection=[id, name, isActive, get_field(structField@")`, with the existing tail match unchanged?

Same thought on the `", field2) > 30"` fragment just below (and on line 403): anchoring it to `predicate=` would stop it also matching inside `pruning_predicate=`.

crates/datafusion/tests/plan_tests.rs
```diff
@@ -148,26 +148,37 @@ async fn verify_plan(
...
+            file_group_noun(*planned_input_partitioned)
+        )),
+        "Plan should scan {planned_input_partitioned} file group(s)"
     );
 }
 
+/// How the plan spells the file-group count: `hoodie.read.input.partitions`
+/// decides how many groups the scan is split into, and DataFusion renders one
+/// group in the singular.
+fn file_group_noun(count: i32) -> &'static str {
```
linliu-code: nit: this helper can go — `"file_groups={2 group"` is already a prefix of `"file_groups={2 groups:"`, so `plan.contains(&format!("file_groups={{{planned_input_partitioned} group"))` covers both the singular and plural rendering on its own.

python/src/datafusion_internal.rs
```diff
@@ -40,23 +40,26 @@ fn extract_codec(session: Bound<PyAny>) -> PyResult<FFI_LogicalExtensionCodec> {
...
+    // the cast below sound. An unnamed capsule no longer passes, where the
+    // previous name-then-read pair let one through.
+    let codec = capsule
+        .pointer_checked(Some(c"datafusion_logical_extension_codec"))
+        .map_err(|e| {
+            PyValueError::new_err(format!(
+                "Expected a PyCapsule named 'datafusion_logical_extension_codec': {e}"
+            ))
+        })?;
+    // SAFETY: DataFusion puts that name only on a capsule holding an
```
linliu-code, line 54: The name check proves the capsule holds an `FFI_LogicalExtensionCodec`, but not one of a matching major version — the capsule name is not versioned, and `datafusion-ffi = "54.1"` is a caret req while the `datafusion==54.0.0` pin lives only in an optional extra. A user who installs a datafusion-python built on ffi 55 into the same env hands over the same capsule name with a different struct layout, and the cast then reads function pointers at the wrong offsets. `FFI_LogicalExtensionCodec` carries a `version` fn returning the major version for exactly this handshake — worth calling it, or softening the comment to say the version match is assumed? (Pre-existing, not introduced here.)

linliu-code (approved):
Thanks for working on this! Six coordinated upgrades in one PR with the churn kept mechanical, and the two genuinely non-obvious pieces — telling a HEAD from a GET by the `head` flag now that object_store routes both through `get_opts`, and rewriting the plan assertions for DF 54's pushdown — are both explained in the description rather than left for a reviewer to reconstruct.

I checked the pieces that could quietly change behaviour and they hold up: `ObjectStoreExt::head` really is `get_opts(head: true)` and every cloud client maps that back to `Method::HEAD`, so the counting store's attribution is unchanged (the exact `(gets, heads)` assertions in hfile.rs and log_file/reader.rs still pass untouched); `copy` and `copy_if_not_exists` both collapse into `copy_opts`, so forwarding it covers both; `statistics()` is gone from DF 54's `ExecutionPlan` and `partition_statistics(None)` walks the same partitions the deleted helper did; `set_max_row_group_row_count(Some(n))` is literally what the deprecated setter did, with `max_row_group_bytes` still defaulting to `None`; and the lance 4→11 jump keeps reading the committed v9_lance fixtures, which is the compat question I most wanted answered.

One thing outside the diff, so I could not leave it inline: `README.md` line 365 still says `cargo add tokio@1 datafusion@52` in the DataFusion quickstart. A user following it would get datafusion 52 (arrow 57) against a crate on 54 (arrow 58), so `register_table` would not compile. It looks like it was missed alongside the demo-app pins, which makes sense — it is the one copy of this version that CI does not build.

The rest of the comments are about the plan_tests assertions, which came out weaker than what they replaced. None of it is blocking, so approving — the README pin is worth picking up before merge, and the other three can be taken or left.

crates/datafusion/tests/plan_tests.rs
yihua: Done. Both anchored: the projection assertion now matches `projection=[id, name, isActive, get_field(structField@`, so a scan that stopped projecting fails it, and the filter fragment is checked against a `scan_predicate` helper that slices out `, predicate=` and cuts before `, pruning_predicate=`. Applied to the v8 assertion too.

crates/datafusion/tests/plan_tests.rs
yihua: Done, helper dropped. `file_groups={N group` covers both renderings on its own.

python/src/datafusion_internal.rs
yihua: Good catch, the comment overstated it. Softened rather than adding a check: `version` sits near the end of `FFI_LogicalExtensionCodec`, so reading it already assumes the layout it would be verifying, and datafusion-ffi itself only calls it in its own module-loading tests. The comment now says the name settles what the capsule holds but not which `datafusion-ffi` built it, and names the pinned `datafusion` extra as what keeps them in step.

Issue #286, opened 2025-02-01
Handle partition filter properly for timestamp based and custom keygen


xushiyan:
Partition path can be in forms that do not match partition filter.

yunchipang:
@xushiyan do we want something like:
1. user can input filters, e.g.`year=2025` or `month=01`
2. transform the filters to timestamp-based key, e.g. `ts >= xxxxx AND ts < xxxxx` and apply it when filtering partitions

not sure how to handle custom keygens, but i'd love to start with fixing timestamp-based keygens.

PR #724, opened 2026-09-04
fix(core): correct hive-style partition pruning for timestamp key generators

### LINKED ISSUE #798: Hive-style partition pruning with TimestampBasedKeyGenerator excludes every partition
### Is there an existing issue for this?

- [x] I have searched the existing issues — #286 (closed, added the transform via #462), #580 (multi-field ordering), #549 (CustomKeyGenerator) and #779 (storage listing) are adjacent; none covers the hive-style path shape.

### Description of the bug

On a hive-style table with `hoodie.table.partition.fields=ts` and `hoodie.keygen.timebased.output.dateformat=yyyy-MM-dd`, an equality predicate on `ts` returns zero file slices and zero rows although a partition holds a matching row. No warning is logged. Without the `hoodie.keygen.timebased.*` keys the same predicate returns the right row, because the pruner logs `Failed to create TimestampBasedKeyGenerator ... Filters will not be transformed` and column-stats pruning takes over. Supplying the generator's parameters turns a correct read into a wrong one.

Cause, at `release-0.5.0`: [`parse_partition_fields`](https://github.com/apache/hudi-rs/blob/release-0.5.0/crates/core/src/keygen/timestamp_based.rs#L281-L302) derives one hive field name per `/`-separated token of `output.dateformat` (`yyyy` → `year`, …), and an unrecognised token becomes its own name, so a format with no `/` yields the field name `yyyy-MM-dd`. [`format_partition_path`](https://github.com/apache/hudi-rs/blob/release-0.5.0/crates/core/src/keygen/timestamp_based.rs#L425-L444) then rewrites the predicate to `_hoodie_partition_path = "yyyy-MM-dd=2024-03-02"`. The directory is `ts=2024-03-02`. Hudi's `TimestampBasedAvroKeyGenerator.getPartitionPath` prefixes the whole formatted value with `partition.fields[0] + "="` exactly once, so no hive-style timestamp-keygen table can match the rewritten path. The shape does not depend on `timestamp.type`.

#724 corrects the path shape and carries Hudi 0.15 and 1.2 fixtures for it; this issue is the symptom-level entry for that PR.

### Steps To Reproduce

A COW table, version 6, three day partitions, produced by a non-Hudi writer (so a second implementation, independent of the fixtures in #724):

```
ts=2024-03-01/<one base file, 1 row>
ts=2024-03-02/<one base file, 2 rows, one with ts = 2024-03-02 12:00:00>
ts=2024-03-03/<one base file, 3 rows>
```

`hoodie.properties` — the writer persists the first three lines; the four `hoodie.keygen.timebased.*` lines were added by hand so the generator could be built (they are the keys Hudi 1.x persists on its own; the shipped `v9_timebasedkeygen_*` fixtures carry `timestamp.type` and `output.dateformat`):

```
hoodie.table.partition.fields=ts
hoodie.datasource.write.hive_style_partitioning=true
hoodie.table.keygenerator.class=org.apache.hudi.keygen.TimestampBasedKeyGenerator
hoodie.keygen.timebased.timestamp.type=DATE_STRING
hoodie.keygen.timebased.input.dateformat=yyyy-MM-dd HH:mm:ss
hoodie.keygen.timebased.output.dateformat=yyyy-MM-dd
hoodie.keygen.timebased.timezone=UTC
```

```rust
let opts = ReadOptions::new().with_filters([("ts", "=", "2024-03-02 12:00:00")])?;
table.get_file_slices(&opts).await?   // 0 slices
table.read(&opts).await?              // 0 rows
```

Remove the four `hoodie.keygen.timebased.*` lines and the same call returns 1 slice and 1 row.

### Expected behavior

1 file slice (`ts=2024-03-02`) and 1 row, the same as the read without the keygen keys.

### Screenshots / Logs

With the keys present, `RUST_LOG=warn` prints nothing. Without them:

```
WARN hudi_core::table::partition] Failed to create TimestampBasedKeyGenerator: Config error: Config 'Missing required configuration: hoodie.keygen.timebased.timestamp.type' not found. Filters will not be transformed.
```

### Software information

- Operating system: macOS 27.0, arm64
- Rust version: rustc 1.98.1
- Project version: hudi 0.5.0 (crates.io, `default-features = false`)

### Additional context

A date-only literal on the same table, `("ts", "=", "2024-03-02")`, fails separately with `Failed to parse date string '2024-03-02' with format '%Y-%m-%d %H:%M:%S': premature end of input`, because the literal is parsed with the column's `input.dateformat`. That is #746's case, not this one.

slachiewicz:
## Description

Closes #798.

Hive-style partition pruning has never worked for `TimestampBasedKeyGenerator`, and two
separate defects made it fail in opposite directions. Every timestamp-keygen fixture in
`crates/test/data` is `hive_style_partitioning=false`, so the hive branch shipped with no
table-backed coverage and both survived #286.

**The path shape was invented.** `format_partition_path` split `output.dateformat` on `/`
and derived a column name per token (`yyyy` to `year`, `MM` to `month`), so field `ts` with
format `yyyy/MM/dd` produced `year=2024/month=03/day=01`. With no `/` in the format it used
the format string itself as the name, giving `yyyy-MM-dd=2024-03-01`. Hudi does neither. The
final line of `TimestampBasedAvroKeyGenerator.getPartitionPath` is:

```java
return hiveStylePartitioning ? getPartitionPathFields().get(0) + "=" + partitionPath : partitionPath;
```

The whole formatted value gets one prefix, taken from `hoodie.table.partition.fields`. Any
`/` inside the format stays in the value and becomes nested directories. So the real paths
are `ts=2024/03/01` and `ts=2024-03-01`, and the reader's predicate matched neither. Where
the writer configs were present this excluded every partition; where they were absent the
filter failed to bind and the table was scanned in full.

**Negation dropped whole partitions.** The transform is lossy: a partition holds every
instant of its period, so `ts != 2024-03-01T14:30:00Z` is satisfied by almost every row of
`ts=2024-03-01`. Mapping the negation onto the path discarded that partition and lost all of
them, which is a wrong-results bug rather than a slow one. `!=` and `NOT IN` now emit no
partition predicate and are left to row-level evaluation.

Four fixtures come with this: three written by Hudi 0.15.0 (table version 6, the last
release that writes it) and one by Hudi 1.2.0 on Spark 4.1 (version 9), all COW. They cover
day granularity, an hour format whose slashes nest under a single hive prefix, and a custom
key generator pairing an identity field with a time-derived one. The two custom-keygen
tables are deliberately a pair, and they document something a reader has to live with:
version 6 records `hoodie.table.partition.fields=region,ts_str` and persists no per-field
key types at all, while version 9 records `region:SIMPLE,ts_str:TIMESTAMP`.

One expectation shifted while writing the integration tests, and it is worth stating rather
than leaving for a reviewer to rediscover. Partition pruning widens a half-open upper bound
to the boundary partition, since the transform has already discarded the time of day.
File-level column statistics then drop that file when its real range misses the predicate.
The two stages compose and the observable result is the tighter of the two.

`CustomKeyGenerator` support for #549 is not in this PR. The version 6 fixture showed that
the obvious fallback for an untyped field is unsafe, and that needs its own change.

Verified: `make format check test` passes, and `cargo clippy --all-targets --all-features
--workspace -- -D warnings` is clean.

- [ ] N/A
- [ ] Manual tests


yihua (commented):
Thanks for working on this! The path-shape correction faithfully mirrors `TimestampBasedAvroKeyGenerator.getPartitionPath` — one hive prefix from the declared partition field over the whole formatted value — and taking `Ne`/`NotIn` out of partition pruning closes a real wrong-results hazard; the unit tests pin both well, and the fixtures accurately capture what 0.15 and 1.2 writers persist.

The inline comments center on one theme I verified locally: the fixture-backed integration tests pass with the fix reverted, because the v6 fixtures don't persist `timestamp.type` so the transform never engages, and when it does engage (as on 1.x tables) the query fails on the un-unescaped `input.dateformat` from `hoodie.properties`. So the end-to-end coverage this PR describes isn't in place yet, and the escaping issue needs a resolution or a tracked follow-up for the fix to be reachable on real tables. Also flagged the two custom-keygen fixtures that no test currently reads.

This is one review pass, not an approval.

> [!NOTE]
> This review was prepared with AI assistance (Claude). The revert/option experiments described in the comments were run locally against this PR's head, but please apply the usual scrutiny to the reasoning and conclusions.

crates/core/tests/table_read_tests.rs
```diff
@@ -3477,3 +3477,143 @@ mod incremental_window_boundaries {
         Ok(())
     }
 }
+
+mod hive_style_keygen_partition_pruning {
```
yihua, line 3481: I verified locally that all four tests in this module pass with the keygen fix reverted to main: the v6 fixtures' hoodie.properties don't carry `hoodie.keygen.timebased.timestamp.type` (0.15 sets it at write time but doesn't persist it), so `TimestampBasedKeyGenerator::from_configs` errors, the warn branch in partition.rs skips the transform, and these assertions are satisfied by file-level column stats alone. Could you make the tests engage the transform (e.g. `Table::new_with_options` with timestamp.type — which currently surfaces the escaping issue I raised separately) or rewrite the comments to describe what is actually pruning here? As written, e.g. "Before the reader built hive paths correctly this returned every file instead" doesn't hold for this fixture.

crates/core/src/keygen/timestamp_based.rs
```diff
@@ -509,7 +482,13 @@ impl KeyGeneratorFilterTransformer for TimestampBasedKeyGenerator {
...
         match filter.operator {
-            ExprOperator::Eq | ExprOperator::Ne => {
+            // A negated predicate cannot prune, because this transform is lossy: a
+            // partition holds every instant of its period, so `ts != 2024-03-01T14:30:00Z`
+            // is satisfied by almost every row of `ts=2024-03-01`. Mapping the negation
+            // onto the path would drop that whole partition and lose all of them. Emit no
+            // partition filter and let the predicate be enforced per row.
+            ExprOperator::Ne | ExprOperator::NotIn => Ok(vec![]),
+            ExprOperator::Eq => {
                 let dt = self.parse_timestamp(&filter.values[0])?;
```
yihua, line 492: When the transform does engage (1.x tables persist timestamp.type), a filtered read currently fails outright: on the existing v9_timebasedkeygen_nonhivestyle fixture, `ts_str >= '2023-11-14T00:00:00.000Z'` errors with TimestampParsingError because `parse_data_for_options` keeps the Java-properties escaping, so input.dateformat arrives as `yyyy-MM-dd'T'HH\:mm\:ss.SSSZ`, and hoodie.properties overrides any user-supplied value. That means the hive-style pruning fixed here has no reachable end-to-end path yet — v6 tables skip the transform, v8/v9 tables error on it. Would you consider unescaping in `parse_data_for_options` (java.util.Properties semantics) in this PR, or filing a linked follow-up? The integration tests can only pin the fixed behavior once this is resolved.

crates/core/tests/gold_parity_tests.rs
```diff
@@ -85,6 +85,13 @@ const NULL_TOKEN: &str = "<null>";
 /// being compared.
 const EXPECTED_WITHOUT_GOLD: &[&str] = &[
     "table_hfile_log_block [MorAvro]",
+    // Partition-pruning fixtures. They exist to pin the on-disk partition path shape a
+    // hive-style timestamp or custom key generator produces, which is asserted directly in
+    // table_read_tests, so they carry no Hudi read snapshot.
+    "v6_customkeygen_hivestyle [Cow]",
```
yihua, line 91: non-blocking: v6_customkeygen_hivestyle and v9_customkeygen_hivestyle aren't read by any test, so the "asserted directly in table_read_tests" part of this comment only holds for the two timebased fixtures. Could you either add a small smoke test pinning the custom-keygen layout (the v6 one is readable today) or ship these two fixtures with the #549 change, and adjust the comment either way?

crates/core/tests/table_read_tests.rs
slachiewicz: Addressed in 76026c0: the three v6 hive-style tests now open the table with `Table::new_with_options(..., hoodie.keygen.timebased.timestamp.type=DATE_STRING)` so the keygen transform engages, and the comments were rewritten to state that explicitly (column-stats-only pruning is now documented as covered by `custom_keygen_partition_paths`). Also added `v9_table_prunes_through_its_persisted_keygen_config` which exercises the transform from persisted table config alone. All rust tests pass on this head.

crates/core/src/keygen/timestamp_based.rs
slachiewicz: Addressed in cc343d8: `parse_data_for_options` in `crates/core/src/config/util.rs` now applies Java-properties unescaping (`\\:`, `\\=`, `\\ `, `\t/\n/\r/\f`, `\uXXXX` with surrogate-pair recombination, malformed escapes degrade to drop-the-backslash). This fixes the `input.dateformat=yyyy-MM-dd'T'HH\\:mm\\:ss.SSSZ` -> `TimestampParsingError` failure on 1.x tables, and `v9_table_prunes_through_its_persisted_keygen_config` (added in 76026c0) pins the end-to-end pruned read on the v9 fixture. Also strips `:SIMPLE/:TIMESTAMP` annotations from `partition.fields` in `table.rs`.

crates/core/tests/gold_parity_tests.rs
slachiewicz: Addressed in 76026c0: added `mod custom_keygen_partition_paths` in `table_read_tests.rs` with `identity_field_prunes_like_a_plain_partition_column` (v6 fixture) and `annotated_partition_field_names_still_resolve_v9` (v9 fixture, pins the `:SIMPLE/:TIMESTAMP` stripping). The `EXPECTED_WITHOUT_GOLD` comment you flagged now holds for all four fixtures.

Issue #609, opened 2026-05-12
refactor(core): use typed values internally in Filter API


xushiyan:
## Summary

`hudi_core::expr::filter::Filter` represents filter values as `Vec<String>`:

```rust
pub struct Filter {
    field: String,
    operator: ExprOperator,
    values: Vec<String>,
}
```

This forces every typed value (Arrow `ScalarValue`, DataFusion literal, decimal, timestamp, binary, …) to be stringified at the boundary and parsed back to a typed `Scalar<ArrayRef>` at apply time via `arrow_cast::cast_with_options` (`crates/core/src/expr/filter.rs:345`). The round trip is lossy: any disagreement between the producer's `Display` and `arrow_cast`'s parser becomes a silent correctness bug — the filter applies but never matches, returning zero rows.

This proposal: make `Filter::values` typed internally while preserving the string-tuple API for cross-language callers (Python, SQL `CREATE EXTERNAL TABLE` options).

## Current state

`Filter` is constructed three ways, all string-flavored:

- `Filter::new(field: String, op: ExprOperator, values: Vec<String>)` — `crates/core/src/expr/filter.rs:38`
- Builder via `FilterField::eq/ne/lt/...(value: impl Into<String>)` — `filter.rs:237-285`
- Tuple parser `from_str_tuples` for cross-language entry points — `filter.rs:156`

At apply time, values are cast back into the column's Arrow `DataType` for comparison against record batches. The DataFusion integration goes through this round trip on every pushed-down filter (`crates/datafusion/src/util/expr.rs:108-220`).

## Bugs this has already produced + lurking

| Type | DataFusion `Display` | `arrow_cast` expects | Status |
|---|---|---|---|
| `Decimal128/256` | unscaled integer (`\"7500\"`) | scaled decimal (`\"75.00\"`) | **Was broken**; patched in `crates/datafusion/src/util/expr.rs` (`scalar_to_filter_value`) — required ~50 lines of formatting per direction |
| `Timestamp{Nanosecond,Microsecond,Millisecond,Second}` | unit/timezone-dependent | parseable form | Likely lurking; no round-trip tests |
| `Binary` / `FixedSizeBinary` | debug-style `[1, 2, 3]` | hex / base64 (unclear) | Almost certainly broken |
| `Float NaN/Inf` | `\"NaN\"`, `\"inf\"` | parses but `NaN = NaN` is false in SQL | Semantic mismatch |
[...]

PR #746, opened 2026-09-04
fix(core): parse timestamp keygen filter literals by their own spelling

slachiewicz:
## Description

> **Stacked on #725, which is stacked on #724.** The diff shown here includes both PRs'
> commits and will shrink to the four below once they merge. Review the last four commits
> only: `test(core): add a timestamp-column fixture for the timestamp key generator`,
> `fix(core): parse timestamp keygen filter literals by their own spelling`,
> `fix(core): keep reading when a partition filter cannot be rewritten` and
> `fix(datafusion): forward timestamp keygen source-column predicates to listing`.

`TimestampBasedKeyGenerator` parsed an incoming filter literal with
`hoodie.keygen.timebased.timestamp.type`, which describes how the *source column* stores
its value. A literal is spelled by whoever wrote the query, and the two need not agree. A
table whose partition column is a real `TIMESTAMP` can only declare an epoch type honestly,
since that is what Spark hands the key generator, and then no datetime a caller writes
survives `parse::<i64>()`. The failure was not a degraded scan: the error came out of
`PartitionPruner::new` and the read failed outright.

The reachable paths turned out to be narrower than the description above suggests, and the
new fixture pins which is which. A DataFusion timestamp literal never hit the parser at
all, because `ScalarValue`'s `Display` prints the raw epoch integer in the column's unit,
which is exactly the declared encoding; `test_convert_timestamp_literal_is_epoch_integer_in_its_unit`
pins that spelling since the key generator now depends on it. The string filter API is
the path that failed: `ReadOptions::with_filters(("ts", ">=", "2024-03-01T00:00:00Z"))`,
and with it the Python and C++ bindings, returned
`Failed to parse epoch microseconds '2024-03-01T00:00:00Z'` on the fixture.

**Literals are parsed by their own spelling.** The configured encoding is still tried
first, so every input that works today behaves exactly as before, and the fallback takes
RFC 3339 with an offset, a naive datetime with `T` or space, and a bare date, all as UTC.
The two families are disjoint: no epoch integer parses as a datetime and no datetime
parses as an integer, so the fallback cannot reinterpret an accepted input, and
`test_parse_timestamp_literal_families_are_disjoint` holds that. `input.dateformat` keeps
describing the source column and is not consulted for the fallback. One side effect worth
naming: a `DATE_STRING` table with a time-bearing input format now accepts a bare date
literal, which previously errored because the format demanded a time.

**An unparseable literal no longer fails the read.** Pruning is an optimization, so a
filter the generator cannot rewrite is kept untransformed, the way #724 already handles
`!=` and `NOT IN`. It still names the source column, which the partition schema does not
contain, so it fails to bind and lands in #725's `unapplied_filters` while the read goes
ahead and enforces it per row.
`unparseable_literal_is_reported_as_unapplied_instead_of_failing_the_read` checks that
composition.

**DataFusion never forwarded the predicate.** Once the fixture existed, the DataFusion
query returned the right rows but scanned all four partitions. The provider classified
partition predicates by the partition schema, which for a timestamp key generator holds
only `_hoodie_partition_path`, so a predicate on `ts` was never passed to file listing.
`HudiDataSource` now also lists the key generator's source column for listing purposes.
Exactness is unchanged: a predicate on a derived column is never `Exact`, because a day
partition holds every instant of that day, and the test asserts the equality query still
returns one row rather than the whole day.

The fixture is a table version 9 COW written by Hudi 1.2.0 on Spark 3.5.7 through the
DataFrame writer. Two things ruled out the Spark SQL recipe the other fixtures use: Hudi
0.15.0 has no `EPOCHMICROSECONDS` and formats the micros as millis (year 56134), and Hudi
1.2.0's SQL layer re-parses a timestamp partition path as a long in
`SqlKeyGenerator.convertPartitionPathToSqlType` and throws on `2024-03-01`. Both are
recorded in the fixture's recipe.

Not in this PR: `CustomKeyGenerator` (#549), and the typed-value refactor in #609. That
refactor is the proper end state for this class of bug, since a typed `Filter` would carry
the DataFusion literal without a string round trip at all. This change is the narrow
increment ahead of it because the string API is what the bindings expose today and will
keep exposing after #609, whose plan keeps the string tuple form as the cross-language
entry point, so a parser that reads a literal by its own spelling is needed either way.

Verified: `cargo clippy --all-targets --all-features --workspace -- -D warnings` is clean,
the no-spill `hudi-core` leg passes, and the keygen, partition, `table_read_tests` and
`hudi-datafusion` suites pass locally. `make format check test` did not complete locally
(the machine ran out of disk during the Python build), so CI is the check for that.

- [ ] N/A
- [ ] Manual tests


slachiewicz:
Parking this as a draft: it needs more work before it is ready for review, and its base has drifted from the rest of the stack, so it does not yet carry the escaping fix it depends on. Picking it back up next week.

*This comment was created with AI assistance.*

PR #751, opened 2026-09-05
fix(ci): install a modern protoc for the manylinux wheel build

### LINKED ISSUE #750: manylinux wheel build fails: container protoc is too old for lance-encoding
The manylinux wheel build fails because the container's `protoc` is too old, so `0.5.0-rc.1` published to PyPI without any Linux wheel and without an sdist.

`before-script-linux` installs protoc with `yum install -y protobuf-compiler`. On the manylinux2014 base (CentOS 7) that resolves to `protobuf-compiler-2.5.0-8.el7`, and prost-build passes `--experimental_allow_proto3_optional`, which protobuf only understands from 3.12 onward:

```
---> Package protobuf-compiler.x86_64 0:2.5.0-8.el7 will be installed
...
error: failed to run custom build command for `lance-encoding v11.0.0`
  Error: Custom { kind: Other,
    error: "protoc failed: Unknown flag: --experimental_allow_proto3_optional\n" }
```

This became reachable when lance became a hard dependency of hudi-core. The aarch64 step installs protoc the same way and has the same latent problem; it never ran because the x86_64 step fails first.

Regular CI runs on ubuntu-latest, where `apt-get install protobuf-compiler` gives a current protoc, so nothing exercises the manylinux container until a release tag is pushed. By then the crates.io versions are published and cannot be reused, so the whole release has to move to a new RC number.

Two things are needed:

- install a protoc that is new enough, rather than the distro package
- build the wheel in that same container during regular CI, so the break surfaces on the pull request

Reproduced locally against `quay.io/pypa/manylinux2014`: the yum package gives `libprotoc 2.5.0` and rejects the flag; installing an official protoc release gives `libprotoc 36.1` and `lance-encoding v11.0.0` then compiles.

yihua:
## Description

closes #750

The manylinux `before-script-linux` installed protoc with `yum install -y protobuf-compiler`, which on the CentOS 7 base resolves to protoc 2.5.0. That predates `--experimental_allow_proto3_optional`, the flag prost-build passes when compiling lance-encoding's protos, so the publish step fails now that lance is a hard dependency of hudi-core. This installs a pinned, sha256-verified official protoc release instead, from a single script (`.github/scripts/manylinux-build-deps.sh`) shared by both release steps and the CI job, so a version bump cannot land in some call sites and not others. The aarch64 step had the same latent bug and only escaped because the x86_64 step aborts the job first. Getting the container to build also surfaced that `librocksdb-sys`'s bindgen finds no usable libclang on the CentOS 7 base; the script installs the LLVM 7 toolset and exports `LIBCLANG_PATH` and `BINDGEN_EXTRA_CLANG_ARGS`, and is sourced from `before-script-linux` because a step-level `env:` never enters the container.

Regular CI runs on ubuntu-latest, whose apt protoc is current, so nothing exercised the container until a tag was pushed and the crates.io version was already burnt. The new `manylinux-wheel-build` job runs the same maturin-action against the same pinned image with the same dependency script, on both targets (aarch64 on the native ARM runner the python tests already use), differing only in `build` instead of `publish`, so the break shows up on the pull request. Both legs pass on this PR. It is deliberately not path-gated: no lockfile is committed, so dependency versions resolve fresh on every run and a break can arrive with no file in the repo changing.

Verified against `quay.io/pypa/manylinux2014`: the yum package gives `libprotoc 2.5.0` and rejects the flag, while the new script gives `libprotoc 36.1` and `lance-encoding v11.0.0` then compiles cleanly.

- [ ] N/A
- [x] Manual tests

The `manylinux-wheel-build` CI job added here is itself the automated coverage: it fails on this change's `main` parent and passes with it.

Manually, in the release container (`quay.io/pypa/manylinux2014`):

| step | result |
| --- | --- |
| `yum install protobuf-compiler` then `protoc --version` | `libprotoc 2.5.0` |
| that protoc with `--experimental_allow_proto3_optional` | rejected, `Missing value for flag` |
| this change's script, then `protoc --version` | `libprotoc 36.1` |
| `cargo build` of `lance-encoding v11.0.0` | finished, exit 0 |
| `cargo build` of `librocksdb-sys v0.16.0` with the exported paths | finished, exit 0 |

### DIFF
--- a/.github/scripts/manylinux-build-deps.sh
+++ b/.github/scripts/manylinux-build-deps.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# Install the system build dependencies for a manylinux wheel.
+#
+# Both release steps and the CI job that guards them run this one script, so a
+# change cannot land in some of them and not others.
+#
+# protoc: the distro package is 2.5.0 on the CentOS 7 base, which predates the
+# --experimental_allow_proto3_optional flag prost-build passes when it compiles
+# lance-encoding/lance-file's .proto files.
+#
+# libclang: librocksdb-sys generates its bindings with bindgen, which needs a
+# libclang the base image does not carry; the distro clang is 3.4, older than
+# bindgen supports. Callers must point LIBCLANG_PATH at the directory printed
+# at the end, since an exported variable would not survive this subshell.
+#
+# perl-IPC-Cmd: needed by openssl.
+
+set -euo pipefail
+
+PROTOC_VERSION=36.1
+
+case "$(uname -m)" in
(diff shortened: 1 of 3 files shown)


bhasudha (approved):
Lgtm

PR #753, opened 2026-09-05
fix(release): reject an expired signing key in the src release verification

### LINKED ISSUE #752: verify_src_release.sh passes an expired signing key and breaks without a TTY
Two problems in `release/verify_src_release.sh`, both hit while verifying `0.5.0-rc.1`.

**The signature check accepts an expired signing key.** The script runs `gpg --verify "$pub_key" "$src"` under `set -o errexit`. When the signing key has expired, gpg reports the signature as good, prints `Note: This key has expired!`, and **exits 0**, so the check passes. Observed directly against a release signed with a key that had since expired:

```
gpg: Good signature from "..." [expired]
gpg: Note: This key has expired!
[GNUPG:] EXPKEYSIG ...
gpg exit code: 0
```

The ASF release-signing guidance is explicit that this should not count as valid: "A signature is valid, if gpg verifies the .asc as a good signature, and doesn't complain about expired or revoked keys. Technically: `gpg --verify --status-fd 1 ...` should classify the .asc as a GOODSIG." The script should require `GOODSIG` rather than relying on the exit code, so `EXPKEYSIG` and `REVKEYSIG` fail.

**`docker run -it` breaks whenever there is no TTY.** The license check at the end uses `-it`. Run from anything without a terminal (CI, a pipe, a script) it dies with `the input device is not a TTY`, which under `errexit` aborts the script and silently skips both the license check and the binary-file check that follows. The `-t` serves no purpose for a non-interactive container.

yihua:
## Description

closes #752

`gpg --verify` exits 0 for a good signature made by an expired or revoked key, emitting `EXPKEYSIG`/`REVKEYSIG` instead of `GOODSIG`, so the bare call under `errexit` accepted a release whose signing key was no longer valid. ASF's release-signing guidance requires `GOODSIG`, so this matches on `--status-fd` output rather than the exit code.

The license check also ran its container with `docker run -it`. Without a TTY that fails immediately, and under `errexit` that skipped both the license check and the binary-file check after it. The `-t` is not needed for a non-interactive container.

- [x] N/A
- [x] Manual tests

The signature regression was reproduced with a throwaway key created and used to sign under `--faked-system-time`, expiring before the present, so the signature was made while the key was valid:

| check | result |
| --- | --- |
| `gpg --verify` exit code | `0`, the old check passes |
| status line emitted | `[GNUPG:] EXPKEYSIG ...` |
| this change's check | rejects, exits 1 |

The whole script was then run against the real `0.5.0-rc.1` artifacts in the dev repo. It reports `GOODSIG` and passes, and now reaches the license check (533 files, 0 invalid) and the binary-file check, both of which the `-it` failure previously skipped.

### DIFF
--- a/release/verify_src_release.sh
+++ b/release/verify_src_release.sh
@@ -70,7 +70,15 @@ echo "<<< OK"
 echo ">>> Verifying signature..."
 curl -s "https://dist.apache.org/repos/dist/$repo/hudi/KEYS" >"$work_dir/KEYS"
 gpg -q --import "$work_dir/KEYS"
-gpg --verify "$pub_key" "$src"
+# gpg exits 0 for a good signature made by an expired or revoked key, reporting
+# EXPKEYSIG/REVKEYSIG instead of GOODSIG. ASF requires GOODSIG, so match on the
+# status output rather than the exit code.
+# human-readable output goes to stderr and stays visible; status goes to the file
+gpg --verify --status-fd 1 "$pub_key" "$src" >"$work_dir/gpg_status"
+if ! grep -q '^\[GNUPG:\] GOODSIG ' "$work_dir/gpg_status"; then
+  echo "ERROR: signature is not a GOODSIG (expired, revoked, or untrusted key)."
+  exit 1
+fi
 echo "<<< OK"
 
 echo "Un-tarring the source release artifact"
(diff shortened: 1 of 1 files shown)


bhasudha (approved):
Lgtm

PR #755, opened 2026-09-05
fix(ci): publish the release only after every artifact has built

### LINKED ISSUE #754: publish workflow uploads partial releases when one build fails
The publish workflow fans out into four jobs that each publish independently:

```
validate-release-tag
  ├── release-crates          -> publishes to crates.io
  ├── release-pypi-mac        -> publishes to PyPI
  ├── release-pypi-windows    -> publishes to PyPI
  └── release-pypi-manylinux  -> publishes to PyPI
```

Nothing gates one on another, so each uploads as soon as its own build succeeds. When one of them fails, the others have already published, and the release cannot be repaired in place: crates.io does not allow a version to be published twice, so the whole release has to move to a new RC number.

This is what happened with `0.5.0-rc.1` (#750). The manylinux job failed while all three crates had already gone to crates.io and the macOS and Windows wheels had already gone to PyPI. The result is a version that is permanently taken, with a PyPI release that has no Linux wheel and no sdist.

Splitting the workflow into a build phase and a publish phase would contain this: build every wheel and the sdist first, upload them as workflow artifacts, and run the publish jobs only once every build has succeeded. A build failure then costs a re-run rather than a version number.

yihua:
## Description

closes #754

Rebased on `main` now that #751 has merged; this is a single commit against it.

The four publish jobs each hung off `validate-release-tag` and uploaded as soon as their own build succeeded, with nothing gating one on another. A single build failure therefore left the others already published, and crates.io refuses a second publish of the same version, so the release has to move to a new RC number rather than being fixed in place. That is how `0.5.0-rc.1` ended up permanently taken while its PyPI release has no Linux wheel and no sdist.

This splits the workflow in two. `build-mac`, `build-windows` and `build-manylinux` build every wheel plus the sdist and upload them as workflow artifacts. `publish-pypi` and `publish-crates` both list all three in `needs`, so neither starts unless every build succeeded, and `publish-pypi` uploads the artifacts rather than rebuilding. A build failure now costs a re-run instead of a version number.

- [x] N/A
- [ ] Manual tests

The workflow only runs on a pushed release tag, so the publish half cannot be exercised on a pull request without consuming a real version. The build half is the same maturin-action invocation the `manylinux-wheel-build` job from #751 runs on every PR, and the job graph was checked by parsing the workflow:

```
validate-release-tag
  build-mac            needs=validate-release-tag
  build-windows        needs=validate-release-tag
  build-manylinux      needs=validate-release-tag
  publish-pypi         needs=[build-mac, build-windows, build-manylinux]
  publish-crates       needs=[build-mac, build-windows, build-manylinux]
```

Given that, it is worth landing this after the next RC ships rather than before, so the RC is not the first thing to exercise a rewritten publish path.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,34 +36,11 @@ jobs:
             exit 1
           fi
 
-  release-crates:
-    name: Release to crates.io
-    needs: validate-release-tag
-    runs-on: ubuntu-latest
-    strategy:
-      max-parallel: 1
-      matrix:
-        # order matters here as later crates depend on previous ones
-        package:
-          - "hudi-core"
-          - "hudi-datafusion"
-          - "hudi"
-    steps:
-      - uses: actions/checkout@v7
-
-      - name: Install protobuf compiler
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y protobuf-compiler
-
-      - name: cargo publish
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: |
-          cargo publish -p ${{ matrix.package }} --all-features
-
-  release-pypi-mac:
-    name: PyPI release on Mac
+  # Build every artifact before anything is uploaded. crates.io refuses a second
+  # publish of the same version, so a build that fails after a sibling job has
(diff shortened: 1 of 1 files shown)


bhasudha (approved):
Lgtm

PR #780, opened 2026-09-15
ci: run the workflows on every base, and unbreak the demo mc image

linliu-code:
## What this is

Position **1 of 4** in a series porting work from Onehouse's `hudi-rs` fork onto upstream `main`. The series is linear and must land in order: **01 → 02 → 03 → 04**.

Because a PR to `apache/hudi-rs` must be based on a branch in this repository, the later positions cannot be stacked on GitHub. Each is therefore raised against `main` and its diff is cumulative. **This PR is the base of the series — its diff is its own work only** (10 commits, `3abc4d5..04b9235d`).

## Why

Two independent problems, bundled because neither is useful alone on a fork:

1. **The workflows do not run on a fork, or on a PR whose base is not `main`.** `pull_request.branches: [main]` means a PR onto a feature branch produces *zero* workflow runs — which renders as green rather than as "not run". That is the enabling change for everything downstream.
2. **`demo/infra/mc` cannot build.** The `mc` binary is fetched from `dl.min.io`, which now returns **HTTP 410**, so every `integration-tests` job fails at image build. This affects upstream exactly as much as it affects a fork.

## What changed

- Drop `pull_request.branches: [main]` from `ci.yml`, `code.yml`, `pr.yml`, `asf-allowlist-check.yml`.
- Guard actions only `apache/hudi-rs` may take (coverage upload, PyPI/crates.io publish) behind `github.repository`, so they skip on a fork instead of failing.
- `.github/scripts/workspace-version.sh` — print the single `[workspace.package] version` from one authority, replacing a `grep | head -n 1`.
- Disk-budget fixes for `rust-tests`, `python-tests` and the coverage job; `[profile.dev.package."*"] debug = false`.
- A `cpp-ffi` job covering the `cpp` workspace member.
- `demo/infra/mc/Dockerfile` — take `mc` from MinIO's published image instead of the dead download URL.

## Note for upstream reviewers

Parts of this are fork-shaped and deserve a decision rather than a rubber stamp. The `github.repository` guards are no-ops here (the condition is always true upstream); dropping the base-branch filter widens when workflows run; the `cpp-ffi` job is new CI cost. **The unambiguously upstream-relevant piece is the `mc` Dockerfile fix** — that is broken on `main` today. I am happy to split this into the `mc` fix alone plus a separate CI discussion if the reviewers prefer.

## Testing

`cargo check --workspace --all-targets` and `cargo fmt --all --check` clean at the source commit. In the fork this position is what first produces check runs at all: 18 checks, 16 passing, 2 skipped, 0 failing.


linliu-code:

PR #783, opened 2026-09-15
feat(jvm-ffi,jni): reader_v2 file-group read as an Arrow C stream

linliu-code:
## What this is

Position **4 of 4** in a series porting work from Onehouse's `hudi-rs` fork onto upstream `main`. The series is linear and must land in order: **01 → 02 → 03 → 04**.

A PR to `apache/hudi-rs` must be based on a branch in this repository, so the series cannot be stacked on GitHub. This branch therefore carries the earlier positions' commits too, and the "Files changed" view is cumulative. **The work unique to this PR is 22 commits, `1c6e09c6..baa08e02`** — please review those.

## What it does

Exports a `reader_v2` file-group read to the JVM as an **Arrow C stream**, plus `libhudi_jni.so` as the carrier.

- `crates/jvm-ffi/` — builds the reader, performs the read, and hands back an `FFI_ArrowArrayStream`.
- `crates/jni/` — the JNI entry points (`readFileGroupInto`, `version`).

## Memory and panic safety

This is the part worth a careful pass, so here is what has already been checked:

- **Every new `extern "C"` entry point is `catch_unwind`-guarded.** A Rust panic unwinding into the JVM is undefined behaviour; all three are wrapped and convert a panic into a thrown `NativeReaderException` or an error return.
- **The Arrow C Data Interface contract holds.** `release` is always set, is reclaimed exactly once and is idempotent; `private_data` is deliberately leaked in `new` and reclaimed in `release`. `out_stream` is null-checked before the read and the `ptr::write` is the last statement, so the caller's struct is untouched on every error path.
- **JNI local references are bounded** — each element read runs inside a `with_local_frame`, which pops on both the success and the error path. `throw` checks for a pending exception before raising a second.

## Known

- **arrow-rs's own stream callbacks are unguarded.** `get_schema` / `get_next` / `get_last_error` / `release_stream` carry no `catch_unwind`. This is safe here only because the read is fully materialised into one `RecordBatch` *before* the stream is constructed, so all panic-capable decode happens inside the guarded entry point. That property is load-bearing; a lazy variant would need a per-`next()` guard.
- **`crates/jni` has no tests in-repo** — it is `crate-type = ["cdylib"]`, so `cargo test` compiles it and runs nothing. It is exercised by a JVM consumer outside this repository. Inert by design rather than by accident, but worth knowing.
- The reader context is built with several fields hardcoded, which suits the metadata table but is not obvious from the export's general-sounding name.


linliu-code:
## Do not review yet — this does not compile

Flagging a defect I introduced, before anyone spends time on it.

This branch replaces `cpp/src/lib.rs` with the ported bridge, which exports `new_file_group_reader_with_context`. It does **not** carry a matching change to the demo, which still calls the previous name:

```
demo/apps/hudi-file-group-api/cpp/main.cpp:142:34: error:
'new_file_group_reader_with_options' was not declared in this scope;
did you mean 'new_file_group_reader_with_context'?
```

`main` is self-consistent today — it exports `new_file_group_reader_with_options` and the demo calls it. The port renames the entry point and leaves the caller behind, so `integration-tests (hudi-file-group-api/cpp)` fails to build.

**Scope:** this affects this PR and the one stacked after it. The two earlier positions in the series (#780, #781) do not touch `cpp/src/lib.rs` and are unaffected.

**How it got here:** the branch was assembled by cherry-pick and pushed without a local build — the crate does not currently build in my environment, and CI had not run here yet, so nothing caught it. That is on me, not on the original author.

**What happens next:** the same failure exists on the internal PR this was ported from, and is being fixed there. I will re-port once that lands rather than invent a rename here that diverges from the internal series. Staying in draft until then.

linliu-code:

PR #784, opened 2026-09-15
fix(core): Prune Partition-Directory descent by filters during storage listing

### LINKED ISSUE #779: [FileSystem Listing] When the metadata table is not enabled, Storage-listing partition path discovery lists the entire partition tree instead of pruning by filters.
## Component

`crates/core/src/table/listing.rs` (`FileLister::list_relevant_partition_paths`), `crates/core/src/storage/mod.rs` (`get_leaf_dirs`), `crates/core/src/table/partition.rs` (`PartitionPruner`)

## Summary

When `hoodie.metadata.enable` is off (or the metadata table isn't otherwise available — e.g. `hoodie.table.version <= 6` predates it), `FileLister::list_relevant_partition_paths` falls back to listing partition directories directly from storage.

That fallback currently does a **full recursive listing of every partition directory** under the table's base path, regardless of any partition filters supplied by the query, and only applies `PartitionPruner::should_include` *after* every leaf partition path has already been discovered (`crates/core/src/table/listing.rs`, current `list_relevant_partition_paths`):

```rust
async fn list_relevant_partition_paths(&self) -> Result<Vec<String>> {
    if !is_table_partitioned(&self.hudi_configs)? {
        return Ok(vec![EMPTY_PARTITION_PATH.to_string()]);
    }

    let top_level_dirs: Vec<String> = self
        .storage
        .list_dirs(None)
        .await?
        .into_iter()
        .filter(|dir| !LAKE_FORMAT_METADATA_DIRS.contains(&dir.as_str()))
        .collect();

    let mut partition_paths = Vec::new();
    for dir in top_level_dirs {
        partition_paths.extend(get_leaf_dirs(&self.storage, Some(&dir)).await?);
    }

    // should_include is only applied here, after the full tree is already listed —
    // no filter narrows top_level_dirs or the get_leaf_dirs recursion above it.
    if partition_paths.is_empty() || self.partition_pruner.is_empty() {
        return Ok(partition_paths);
    }

    Ok(partition_paths
        .into_iter()
        .filter(|path_str| self.partition_pruner.should_include(path_str))
        .collect())
}
```

`get_leaf_dirs` (`crates/core/src/storage/mod.rs`) recurses into every directory unconditionally — it takes no predicate parameter at all, so there is nothing to check before calling `storage.list_dirs(subdir)` on a candidate directory. Every directory in the partition tree incurs a storage `list` call (a network round trip on cloud object stores or HDFS), even when its name already violates a supplied filter and could never contain a matching row.

## Impact

For a table with N partitions and a highly selective query (e.g. filtering down to a single partition), this results in O(N) storage list calls per query instead of O(1) or O(depth). On cloud storage (S3/GCS) or HDFS, each `list` call is a network round trip, so this scales query latency and API cost directly with total partition count rather than with the query's actual selectivity — independent of whether the query is otherwise cheap. This is worse on tables with many partitions and only gets worse as the table grows, even though the query itself doesn't.

## Suggested fix

Reject a directory *before* listing its children whenever a supplied partition filter already rules it out, so `list_dirs` is only ever called on the subtree that could possibly contain a match.

## Cross-reference

The Java implementation already does something like this at the listing layer. In `apache/hudi`, `hudi-common/src/main/java/org/apache/hudi/metadata/FileSystemBackedTableMetadata.java`, `getPartitionPathWithPathPrefixUsingFilterExpression` prunes directories level-by-level during its recursive walk, using a partially-bound expression before descending and a fully-bound one on completed partition paths:

```java
// evaluated against each directory before recursing into it
pathsToList.addAll(result.stream()
    .filter(entry -> entry.getValue().isPresent())
    .map(entry -> entry.getValue().get())
    .filter(path -> partialBoundExpr instanceof Predicates.TrueExpression
        || (Boolean) partialBoundExpr.eval(extractPartitionValues(...)))
    .collect(Collectors.toList()));
```

Same two-tier shape as the fix proposed above (partial predicate during descent, full predicate at the leaf)

hakunamatata-sb:
## Description

When the metadata table is unavailable, `FileLister::list_relevant_partition_paths` fell back to a full recursive listing of every partition directory (`get_leaf_dirs`), applying `PartitionPruner::should_include` only after every leaf partition path had already been discovered. A selective query therefore issued as many storage `list` calls as an unfiltered one, scaling cost and latency with total partition count rather than query selectivity.

This change:

- Threads a `should_descend` predicate through `get_leaf_dirs`, checked before recursing into each child directory.
- Adds `PartitionPruner::should_include_prefix`, a partial-path predicate used as that descent-time filter. It fails open (never wrongly prunes) whenever it can't positively prove a resolved segment violates a filter — no filters present, an opaque single-field partition schema (the case for timestamp-based-keygen tables, which already collapse to one `_hoodie_partition_path` field), a malformed hive-style segment, or a cast/comparison error.
- Keeps the existing full-path `should_include` filter unchanged as the final correctness backstop — descent pruning is purely an optimization.

▎ Breaking: storage::get_leaf_dirs gains a required should_descend: &(dyn Fn(&str) -> bool + Sync) parameter to support descent-time pruning. No in-repo caller outside hudi-core uses it; downstream consumers passing a fixed argument list will need to add &|_: &str| true to preserve the old behavior.

Fixes #779

NA

- [ ] N/A
- [x] Manual tests

`cargo test -p hudi-core table::partition::tests` — 23/23 pass, including 6 new unit tests for `should_include_prefix` (hive-style, non-hive-style, url-encoded, malformed segment, no filters, single-field/timestamp-keygen schema).

`cargo test -p hudi-core table::listing::test` — new regression test `partition_filter_pruning_reduces_storage_list_calls` asserts exact storage `list` call counts against a `CountingObjectStore`: 7 calls unfiltered vs. 3 calls with a selective filter (`byteField = "10"`) on `V6ComplexkeygenHivestyle`, plus correctness of the returned partition paths.

`cargo test -p hudi-core storage::tests::use_storage_to_get_leaf_dirs*` still pass with the new predicate parameter.

`cargo clippy -p hudi-core --lib -- -D warnings` clean on the touched files.


crates/core/src/table/partition.rs
```diff
@@ -174,6 +174,71 @@ impl PartitionPruner {
...
+    ///
+    /// Unlike `should_include`, `partial_path` may resolve fewer segments than
+    /// the schema has fields (we're mid-descent, not at a leaf yet). Any segment
+    /// this can't safely reason about — not enough schema fields to compare
+    /// against, a single opaque path field (e.g. under a timestamp-based key
+    /// generator, where a segment carries no independently meaningful value),
+    /// a hive-style segment that doesn't parse as `key=value`, or a cast/compare
+    /// error — is treated as unconstrained, never as a reason to reject. Only a
+    /// segment that positively fails an already-resolved filter returns `false`.
+    pub fn should_include_prefix(&self, partial_path: &str) -> bool {
```
codope, line 204: I think this can drop a partition that actually matches. Take `region=us%2Feast/dt=2024-01-01` with a filter `region = 'us/east'`: the prefix check comes back false, but `should_include` on the full path comes back true. Values with a `/` in them are not that rare. Am I reading it right?

I suspect the main reason is that `parse_segments` decodes the whole path and then splits on `/`. Java goes the other way, splitting first and unescaping each fragment (`AbstractHoodieTableMetadata.extractPartitionValues`). Would it make sense to fix that ordering first? Looks like it would help `should_include` too.

Once segments decode independently, i would suggest gating this on `is_hive_style && is_url_encoded`? That is what Java requires (`FileSystemBackedTableMetadata.java:157`), and it seems to be the thing that makes the positional binding safe.

crates/core/src/table/partition.rs
```diff
@@ -174,6 +174,71 @@ impl PartitionPruner {
...
+        if self.and_filters.is_empty() || self.schema.fields().len() <= 1 {
```
codope: This also switches the optimization off for single-column `dt=...` tables, which is probably the most common layout out there. I tried it on `V6SimplekeygenHivestyleNoMetafields` and got 4 list calls either way. Keying on the `_hoodie_partition_path` field instead of the field count got it down to 2, with the suite still green. Worth a try?

crates/core/src/table/partition.rs
```diff
@@ -174,6 +174,71 @@ impl PartitionPruner {
...
+            return true;
+        }
+
+        let decoded;
```
codope: minor: `let decoded; ... &decoded` reads a bit awkwardly. `Cow<str>` would do this directly.

crates/core/src/table/partition.rs
```diff
@@ -401,6 +466,113 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_partition_pruner_should_include_prefix_url_encoded() {
+        let schema = create_test_schema();
+        let configs = create_hudi_configs(true, true);
+        let filter_eq_a = Filter::try_from(("category", "=", "A")).unwrap();
+        let pruner = PartitionPruner::new(&[filter_eq_a], &schema, &configs).unwrap();
+
+        assert!(!pruner.should_include_prefix("date%3D2023-02-01%2Fcategory%3DB"));
```
codope: I don't think Hudi writes paths like this. It escapes values but leaves `=` and `/` alone. Something like `category=A%2FB` would be closer, and it would cover the slash case too.

crates/core/src/storage/mod.rs
```diff
@@ -409,8 +409,16 @@ impl Storage {
 /// - /usr/hudi/table_name/dt=2025/month=02
 ///
 /// the result is \[".hoodie", "dt=2024/mont=01/day=01", "dt=2025/month=02"\]
+///
+/// `should_descend` is checked before recursing into each child directory, so a
+/// directory a filter already rules out never incurs the storage `list` call
+/// its own children would otherwise need.
 #[async_recursion]
-pub async fn get_leaf_dirs(storage: &Storage, subdir: Option<&str>) -> Result<Vec<String>> {
+pub async fn get_leaf_dirs(
```
codope, line 417: Just noting this is `pub`, so the added parameter is a signature break. Nothing outside the crate calls it, so probably fine. Maybe worth a line in the description?

crates/core/src/table/listing.rs
```diff
@@ -185,9 +185,14 @@ impl FileLister {
             .filter(|dir| !LAKE_FORMAT_METADATA_DIRS.contains(&dir.as_str()))
             .collect();
 
+        let should_descend = |prefix: &str| self.partition_pruner.should_include_prefix(prefix);
+
         let mut partition_paths = Vec::new();
         for dir in top_level_dirs {
-            partition_paths.extend(get_leaf_dirs(&self.storage, Some(&dir)).await?);
+            if !should_descend(&dir) {
```
codope: Could this fold into the `top_level_dirs` filter chain next to the `LAKE_FORMAT_METADATA_DIRS` one, rather than a separate `continue`?

crates/core/src/table/listing.rs
```diff
@@ -311,6 +316,61 @@ mod test {
...
+        assert_eq!(unfiltered_lists, 7);
+
+        let filter_eq_10 = crate::expr::filter::Filter::try_from(("byteField", "=", "10")).unwrap();
+        let selective_pruner =
+            PartitionPruner::new(&[filter_eq_10], &partition_schema, &hudi_configs).unwrap();
+        let (filtered_partitions, filtered_lists) = list_calls_for(selective_pruner).await;
+        assert_eq!(filtered_partitions, 1);
+        assert_eq!(filtered_lists, 3);
+
+        assert!(
```
codope: minor: this cannot fail once the two exact counts above are asserted.

codope (commented):
Thanks for taking this on, and for the `CountingObjectStore` test, that is a nice way to pin the behaviour. The shape of the change looks right to me. One thing I think is a correctness problem (left inline on `should_include_prefix`), plus a few smaller notes.

A couple more things:

- Could you add a test with a `/` in the partition value? And it would be great to see the counting test on a single-level table too, since that is the case the issue is really about.
- The decode/split/cast in `should_include_prefix` mirrors `parse_segments`. Any appetite for pulling out a shared per-segment helper? Easy for the two to drift otherwise.
- `cargo fmt --all -- --check` is unhappy with three spots: the two `get_leaf_dirs` test calls and the `partition_paths.extend(...)` line. The PR title also needs a conventional-commits prefix, since commitlint checks the title. Your commit message already has one.
- Not for this PR, but the walk is serial level by level, while Java lists each level in parallel. On a really wide table I would guess that matters more than the call count.

crates/core/src/storage/mod.rs
hakunamatata-sb: Added the breaking change description in the PR summary, PTAL.

crates/core/src/table/partition.rs
hakunamatata-sb: fixed the single partition issue now that as well get pruned, added a safety check for MetaField:PartitionPath in that case when we would have multiple segments for the single partitions. 
Along with hive-style partition and is_url_encoded to be always ensured. 
The bug related to encoding and splits have been reordered, have done so in the the split_segments and reused that parse_segments so they don't diverge. 
Thanks for pointing this out.

crates/core/src/table/partition.rs
hakunamatata-sb: fixed.

crates/core/src/table/partition.rs
hakunamatata-sb: fixed

crates/core/src/table/partition.rs
hakunamatata-sb: fixed.

crates/core/src/table/listing.rs
hakunamatata-sb: fixed.

crates/core/src/table/listing.rs
hakunamatata-sb: fixed.

hakunamatata-sb:
@codope @xushiyan requesting re-review.

